PR #98, opened 2017-11-10
WIP: Show branch name in deployments

### LINKED ISSUE #85: show ref instead of sha if one exists
https://github.com/github-slack/app/blob/a8cfa539aee3f9b20dd905027d4206cdf9118b8d/lib/slack/renderer/deployment-status.js#L6-L11

---

wilhelmklopp:
Old:
image

New:
image

Fixes #85


davidsteinsland:
Why not use the value of `ref`? This will either contain a SHA, a branch name or a tag. If one wants to output the branch name in slack, users could just set the branch name in the `ref` property when creating the deployment.

scepticulous:
Closing in favor of #915

PR #351, opened 2018-02-23
Flatten structure

bkeepers:
I think the current directory structure is a relic from our first week hacking on this project when we weren't sure what shape this app was going to take. I'd love to flatten it a bit to make the code easier to grok:

- [x] `lib/slack/renderer` => `lib/messages`
- [x] `lib/slack/commands` => `lib/commands`
- [x] `lib/slack/middleware` => `lib/middleware`
- [x] some files in `lib/github` => `lib/activity`
- [x] `lib/slack/unfurls` => `lib/unfurls`
- [x] `test/unit/models` => `test/models`


wilhelmklopp:
>  some files in lib/github => lib/notifications

I think at some point we decided to call "notifications" "activity". Are we sticking with that?

bkeepers:
There's still some stuff I want to move around, but I feel like this is already a lot of changes, so I'll stop there and save the other changes for a future PR.

Issue #164, opened 2017-12-04
Rollup merge/push/deploy events


bkeepers:
Merging a pull request triggers at least 2 notifications (merge, push to master), plus one for each deployment target. I think we should consider rolling these up somehow:

![](https://user-images.githubusercontent.com/173/33563276-18228ee2-d8dd-11e7-9b5a-07813ce80d76.png)

PR #364, opened 2018-02-25
Show full list of commits

### LINKED ISSUE #231: Showing details of commits
Via https://github.com/github-slack/private-beta/issues/4#issuecomment-357059984:

> I don't like how almost no information is shown for pushes. Expecting a bit more meat there (commit message at the very least.)

I agree that we could probably show some more detail on pushes to master, especially if we roll up push/merge events in #164

bkeepers:
Fixes #231, which has been one of the most common pieces of feedback the past few days.

image

For now this just includes all commit messages in the attachment body and relies on Slack to collapse them behind a "Read More" link.


lib/slack/renderer/push.js
```diff
@@ -13,24 +13,32 @@ class Push extends Message {
...
     const text = `<${this.push.compare}|${noOfCommits} new ${commits}> ${pushed} to ${branch}`;
     return {
-      fallback: text,
       text,
     };
   }
 
   getRenderedMessage() {
     return {
+      ...this.getCore(),
```
wilhelmklopp: This "pretext" way of showing the number of commits is very different to how we've been making use of the pretext field so far. It's typically ${github object} ${past tense action} by ${actor}. For example `Pull request merged by bkeepers`.
So instead of showing this message above the attachment, could we move it inside the attachment, maybe as a title or **bolded** first line before list of commits?


If you really want it to stay this way, this line should probably be placed inside the attachment as 
```
...this.getCore()
```

with the return value of this.getCore() being:
```
{
    fallback: text,
    pretext: text,
}
```

lib/slack/renderer/push.js
```diff
@@ -13,24 +13,32 @@ class Push extends Message {
...
+  }
+
   getCore() {
     const noOfCommits = this.push.commits.length;
     const commits = noOfCommits === 1 ? 'commit' : 'commits';
     const branch = `\`${this.push.ref.replace('refs/heads/', '')}\``;
     const pushed = this.push.forced ? 'force-pushed' : 'pushed';
     const text = `<${this.push.compare}|${noOfCommits} new ${commits}> ${pushed} to ${branch}`;
     return {
-      fallback: text,
```
wilhelmklopp: We should definitely keep having a fallback for the push message. Otherwise it'll look really weird in mobile/desktop notifications

lib/slack/renderer/push.js
```diff
@@ -13,24 +13,32 @@ class Push extends Message {
     this.push = push;
   }
 
+  get commits() {
+    return this.push.commits.map((commit) => {
+      const message = commit.message.split('\n')[0];
+      const sha = commit.id.substring(0, 8);
```
wilhelmklopp: does the `sha` really add value to the overall message? It seems like noise, so I would leave it off 😬

bkeepers:
> This is especially true if folks follow the very common workflow of "pushing" to master only when a PR is merged.
> In those cases the title of the PR should be enough context as to what changes have now made it into master.

In this case I think #164 is the solution, where we skip the full push notification entirely and do a summary attachment on the PR notification.

wilhelmklopp:
> In this case I think #164 is the solution, where we skip the full push notification entirely and do a summary attachment on the PR notification.

Oh wow, I almost forgot about that. Definitely want to pick up that idea again. Very good point

lib/slack/renderer/push.js
wilhelmklopp: tried it out: It looks super weird without the sha

lib/messages/push.js
```diff
@@ -13,25 +13,31 @@ class Push extends Message {
...
   getRenderedMessage() {
     return {
       attachments: [
         {
           ...super.getBaseMessage(),
-          ...this.getCore(),
-          mrkdwn_in: ['text'],
+          fallback: this.summary,
+          text: `*${this.summary}*\n${this.commits}`,
+          mrkdwn_in: ['text', 'title'],
```
bkeepers: Nit: doesn't look like `title` is used

Issue #316, opened 2018-02-20
404 when clicking link to install GitHub App


bkeepers:
If you try to subscribe to a repository that doesn't have the GitHub App installed, and you are not an owner of the organization, the "Install GitHub App" link gives you a 404 error.

Steps to reproduce:

1. Given `orgname` is an organization that you are a member of (not an owner), and it does not have the the GitHub App installed
1. type `/github subscribe orgname/repo` for a repository that you have access to
1. Click the "Install GitHub App" link
1. See GitHub 404 page

Current workaround: ask an organization owner to manually [install the GitHub App](https://github.com/apps/slack).

cc @jchuerva @osowskit

a-tal:
+1

this bug breaks all functionality of this integration for us

bkeepers:
@a-tal if you could get someone that is an admin on the organization to [install the app on the organization](http://github.com/apps/slack), then it should work.

a-tal:
@bkeepers while that would work, the workflow there is a little less than ideal. The organization has hundreds of mostly private repos that I would like to hook some (60ish today) up with this integration. Compiling that list, getting someone else to go through the select-a-repo funtimes to find them all, and then maintaining updates to that subset doesn't sound nearly as nice as issuing a slash command in slack.

I don't quite get why org admin is needed if I have repo admin on all the repos I want to enable.

bkeepers:
@a-tal I agree with you. This is feedback that the GitHub Apps team is aware of and is actively working to improve.

chrismessina:
404 isn't the right error code for this kind of issue too. Would be at least helpful to provide the information cited here... i.e. "You don't have sufficient privileges to do that. Please contact your org owner to complete this task."

I was able to install Github on Slack — but apparently don't have the right permissions to get it to actually send data...! Pretty confusing!

PR #367, opened 2018-02-25
Improve message that prompts to install GitHub App

wilhelmklopp:
This improves the message that we post when the github app is not installed for a given repository that a user is trying to subscribe to.

Users have been getting confused as to why the "Install GitHub App" sends them to a page that shows them a 404 error. The reason is because they don't have the necessary permissions to install the app.

With this PR, we do the following:
-  If a user is the owner of a user-owned repo, we show no additional message (just like the current behaviour, because they would not see the 404)
- If a user attempts to subscribe to a user owned repo that they're not the owner of, we suggest that they should ask the 
- If a user attempts to subscribe to an org owner repo, then we add a note saying that they either need to be an org owner to install the app or ask one to do it for them.

Screenshots of the new messages:
Me running `/github-wilhelm subscribe bkeepers/dotenv`:
image

Me running `/github-wilhelm subscribe robotland/test`:
image

Me running `/github-wilhelm wilhelmklopp/wilhelmklopp`:
image

Partially addresses #316


lib/slack/middleware/get-installation.js
```diff
@@ -40,11 +40,13 @@ module.exports = async function getInstallation(req, res, next) {
     const installLink = `${protocol}://${host}/github/install/${owner.id}/${await state.stringify()}`;
 
     if (/api\.slack\.com/.test(req.headers['user-agent'])) {
-      res.json(new InstallGitHubApp(installLink));
-    } else {
-      res.redirect(installLink);
+      if (owner.id === parseInt(gitHubUser.id, 10)) {
+        return res.json(new InstallGitHubApp({ installLink }));
+      }
+      return res.json(new InstallGitHubApp({ installLink, ownerType: owner.type }));
```
bkeepers: Since there's no different in behavior here, you pass owner to the InstallGitHubApp message and do this conditional there? That keeps this code focused on control logic, and the message focused on display logic.

lib/slack/renderer/flow/install-github-app.js
```diff
@@ -1,20 +1,30 @@
...
+    this.ownerType = ownerType;
+  }
+  get guidance() {
+    if (!this.ownerType) {
+      return '\n';
+    }
+    if (this.ownerType === 'User') {
+      return '\n_Note: You will need to ask the owner of the repository to install it for you._';
+    }
+    return '\n_Note: You need to be an organization owner to install the app (or ask one to install it for you)._';
```
bkeepers: > …install it for you

Should this link to the [GitHub app installation page](http://github.com/apps/slack)?

bkeepers:
Also, this is not specific to this PR, but one thing I've been thinking is that this prompt to install the GitHub App isn't very friendly.

We don't necessarily have to solve it now, but at some point I'd love to figure out how to make it more clear what this is and what it means.

lib/slack/renderer/flow/install-github-app.js
wilhelmklopp: I'm not sure I follow on this one. Why would we link to the generic page here?

bkeepers: It's jut not clear what they're asking the owner to do here.  Are they supposed to ask the owner to install the GitHub app? If so, I think we should give them the link to share. Or are they asking them to run the subscribe command?

wilhelmklopp: Ah, that's a very good point! I assumed they would copy the link from the button or from the URL bar, but that's definitely not obvious. We should show the link 👍

PR #371, opened 2018-02-26
Delayed commands

### LINKED ISSUE #352: operation_timeout, but appears to work?
We're getting timeouts when running the migration command, but the command does appear to be working:

screenshot 2018-02-23 13 06 37

The old integration is removed, and the subscription appears in `/github subscribe list`, so it seems like it is actually working?

bkeepers:
Slack has a 3000ms timeout for responding to commands. Since we're so highly dependent on the GitHub API and the Slack API, the response time is often out of our control.

This is an experiment to handle command timeouts automatically. If a command is responded to within the timeout, it will just be returned as an HTTP response. If it takes longer, the command will be acknowledged, and then further responses will be sent to the `response_url` callback.

The key is to just always call `command.respond` and it will do the right thing.

Fixes #352

### DIFF
--- a/lib/middleware/get-installation.js
+++ b/lib/middleware/get-installation.js
@@ -40,11 +40,11 @@ module.exports = async function getInstallation(req, res, next) {
     const installLink = `${protocol}://${host}/github/install/${owner.id}/${await state.stringify()}`;
 
     if (/api\.slack\.com/.test(req.headers['user-agent'])) {
-      res.json(new InstallGitHubApp(installLink));
+      await command.respond(new InstallGitHubApp(installLink));
     } else {
       res.redirect(installLink);
     }
   } catch (err) {
-    res.json(new NotFound(command.args[0]));
+    await command.respond(new NotFound(command.args[0]));
   }
 };
--- a/lib/middleware/route-command.js
+++ b/lib/middleware/route-command.js
@@ -6,7 +6,17 @@ const Command = require('../slack/command');
 module.exports = function route(req, res, next) {
   req.log.debug({ command: req.body }, 'Received slash command');
 
-  const command = new Command(req.body);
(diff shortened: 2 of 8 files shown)


lib/middleware/get-installation.js
```diff
@@ -40,11 +40,11 @@ module.exports = async function getInstallation(req, res, next) {
     const installLink = `${protocol}://${host}/github/install/${owner.id}/${await state.stringify()}`;
 
     if (/api\.slack\.com/.test(req.headers['user-agent'])) {
-      res.json(new InstallGitHubApp(installLink));
+      await command.respond(new InstallGitHubApp(installLink));
```
wilhelmklopp: should we `return` here?

lib/middleware/get-installation.js
```diff
@@ -40,11 +40,11 @@ module.exports = async function getInstallation(req, res, next) {
...
     } else {
       res.redirect(installLink);
     }
   } catch (err) {
-    res.json(new NotFound(command.args[0]));
+    await command.respond(new NotFound(command.args[0]));
```
wilhelmklopp: also here

lib/middleware/route-command.js
```diff
@@ -6,7 +6,37 @@ const Command = require('../slack/command');
...
+    };
+
+    // Slack has a 3000ms timeout, so if we don't respond within that window, ack
+    // the command and use the delayed response URL.
+    timeout = setTimeout(() => {
+      req.log('Command taking too long. Switching to delayed response.');
+
+      // Ack the command
+      command.respond({ response_type: 'in_channel' });
+    }, 2500);
```
wilhelmklopp: I wonder if 2500s may already be too long (because the response has to make its way back to Slack), but i guess we can only try it out and see :)

PR #373, opened 2018-02-26
Fix Twitter card

sophshep:
Realized the social cards weren't working on Twitter (though they work well on Facebook):

image

This _should_ fix them ahead of the comms push tomorrow.


bkeepers:
image

Issue #387, opened 2018-02-27
Subscribing randomly fails with "Could not find resource"


wilhelmklopp:
Experienced this morning:
![image](https://user-images.githubusercontent.com/7718702/36722232-e4bbfc90-1ba4-11e8-9fdd-5bee3e8beec3.png)

Didn't change anything about the state of the installation of `integrations/test`, so this seems kinda weird.

opus131:
Probably related: I see messages being missed randomly as well. I created several test issues and commits, saw only some of them in the subscribed channel while others were not shown.

wilhelmklopp:
@opus131 thanks for confirming.
This seems to be an issue with the GitHub API timing out for a certain input. We're investigating

wilhelmklopp:
Hey folks! I've deployed a change that should stop message deliveries being missed. 

The app will continue to be a little slow until the fix in the GitHub API goes out.

devprodoctor:
We started using Slack yesterday (2018-02-26) and this was already happening. After many many tries, we get a repo to subscribe. All of our repos are private.

wilhelmklopp:
@devprodoctor thanks, that's very useful info! Are you still seeing this problem now?

devprodoctor:
yes

wilhelmklopp:
A fix has gone out for the GitHub API, and we haven't seen this error for the past 15min.

@devprodoctor if you're still having problems, could you open another issue on this repo? 🙇

gorlak:
I've been seeing this again today, with my public repos

gorlak:
Thanks for the quick reply, signin fixed it!! Thanks!

tomgarciap:
> @gorlak would you mind opening a separate issue for this, and if possible add some steps to reproduce it? You could also try doing `/github signin` to see if that fixes it.

Saved my day

PR #388, opened 2018-02-27
Return true and output warning when GitHub returns 500

wilhelmklopp:
Related to the ongoing incident (#387 and https://sentry.io/github-integrations/slack/issues/473125081/)

When GitHub fails with a 500 (or 502) to our check of whether the user still has repo access, we throw an error, which aborts the requests.
This PR modifies that behaviour slightly, so that we log a warning, but still let the request proceed as if we received a 200.

At the very least this PR should temporarily fix the problems we're having with the ongoing incident.


wilhelmklopp:
@bkeepers should we open follow up issues for the very valid points you raise? Either here or in octokit/rest.js

PR #399, opened 2018-02-28
Only prompt to migrate the first time a configuration is imported

### LINKED ISSUE #353: Intro message is sent again if you click "Add to Slack" again
As far as I can tell, clicking "Add to Slack" a second time doesn't actually create a new integration, but it *does* cause the "Good news: ..." message to be resent to every channel that needs migration.

(Also, I am an idiot and clicked the button twice.)

### LINKED ISSUE #396: Track which configurations have been prompted for in-channel and do not reprompt them
We've been sending out upgrade prompts via Slackbot to multiple users. The Github app can be installed multiple times onto Slack, and each time we will send you a list of the active configurations. If you prompt in-channel for each of those configurations for each installation, the channel can easily be overrun with these prompts:

image 2

What we'd like to happen is when you send a prompt into the channel, mark the configuration as "prompted" in your database and check that flag the next time you would prompt and not send the prompt if it's already marked.

bkeepers:
Fixes #396 
Fixes #353

The diff looks like a lot of changes because I had to slightly refactor how the import was being handled and moved it into the model in the process, but the gist is:

`LegacySubscription.import` now only returns _newly_ imported subscriptions, so if a configuration is sent multiple times (which sounds like happens if multiple people get the prompt to install the new GitHub app), we won't prompt multiple times.

### DIFF
--- a/lib/models/legacy-subscription.js
+++ b/lib/models/legacy-subscription.js
@@ -34,32 +34,54 @@ module.exports = (sequelize, DataTypes) => {
     },
   });
 
-  LegacySubscription.import = (configuration) => {
+  LegacySubscription.importAll = async configs => (
+    // Return one flattened list of all the imported subscriptions
+    [].concat(...await Promise.all(configs.map(LegacySubscription.import)))
+  );
+
+  LegacySubscription.import = async (configuration) => {
     if (configuration.repos) {
-      return Promise.all(configuration.repos.map(async (repo) => {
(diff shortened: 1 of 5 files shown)


bkeepers:
Instead of putting this in the user's hands, I think I'd rather see us post a scheduled followup message (~2 weeks later or something) that asks "Hey, it looks like you still haven't activated these subscriptions" with a button that says "Remind me later" and "Don't ask again".

Issue #358, opened 2018-02-23
Error: Can't set headers after they are sent.


bkeepers:
Something in the `/github subscribe` command is attempting to respond to the request twice.

https://sentry.io/github-integrations/slack/issues/469405384/

```
Error: Can't set headers after they are sent.
  File "/app/lib/slack/commands/subscribe.js", line 22, in respondWith
    res.json(message);
  File "/app/lib/slack/commands/subscribe.js", line 52, in module.exports
    return respondWith(new AlreadySubscribed(command.args[0]));
```

wilhelmklopp:
We haven't seen this error in a while, so closing this out. This occurs when we try to send another express response after we've already sent one.

PR #402, opened 2018-02-28
Always return when we invoke respondWith

### LINKED ISSUE #394: subscribing to new feature on existing channel returns noisy message
When subscribing to a new feature on an existing channel, I consistently get:

> Only visible to you
> You're already subscribed to `organization/repository`

wilhelmklopp:
Fixes #394

### DIFF
--- a/lib/commands/subscribe.js
+++ b/lib/commands/subscribe.js
@@ -45,7 +45,7 @@ module.exports = async (req, res) => {
         req.log.debug({ settings }, 'Subscription already exists, updating settings');
         subscription.enable(settings);
         await subscription.save();
-        respondWith(new UpdatedSettings({ subscription, repository: from }));
+        return respondWith(new UpdatedSettings({ subscription, repository: from }));
       }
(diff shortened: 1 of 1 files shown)


bkeepers:
Lol, I was just about to open a PR for this, so I'll paste what I was writing…

#371 exposed a bug where we were forgetting to return after rendering a response, and since we can now call `command.respond` multiple times, it was posting a very confusing message:

slack_-_github___slack

I'd love to write a test for this, but the test was passing since this was technically responding correctly at first (but then was falling through to another response). Since this was missing the `return`/`await`, the test was finishing an no errors were returned.

Ideally we would have a test that detected missing `return`/`await` on `async` functions. You can [do this in typescript](https://www.meziantou.net/2017/11/20/detect-missing-await-in-typescript), but haven't seen anything yet for plain ol Node.

This likely also fixes https://github.com/integrations/slack/issues/358, which hasn't happened since #371 was rolled out.

PR #414, opened 2018-03-02
Store completed unfurls

wilhelmklopp:
This PR introduces a new table, that stores completed unfurls for analytics purposes


wilhelmklopp:
> What would you think about storing the raw event from Slack?

Definitely, but we do some processing after we receive the event, such as figuring out what type of resource we're unfurling (pull, account, etc.), if it's condensed or not, etc. We'd lose that if we just store the raw event from Slack.

> One thing that's been bothering me lately is how a lot of our code has mismatch in shape/structure from data coming from Slack and GitHub (snake case vs camel case, natural IDs vs internal primary keys, etc). I think our life will be a lot easier if we move closer to mirroring the data from the source.

Yeah, totally agree. Staying closer to the source will make life easier 👍 

> In this case, I think it'd be interesting to see the Unfurl record not just as a log of unfurls sent, but as the model that encapsulates all the unfurl logic: parsing the url to understanding what kind of resource it is, checking if it is eligible for unfurl, etc.

At the moment this new table is entirely for analytics purposes, so it's not at all used/doesn't interfere with any of our existing unfurl logic. Are you suggesting that this PR should also refactor all of our unfurl logic so that it moves to the model? I think I'm cool with that, although it'll probably amount to a rather sizeable PR? 
Also I think this current "log" implementation for analytics purposes will still be similar to how it is in this PR so far, just that now it's not the only functionality of the model. 

Getting rid of our use of redis to calculate unfurl eligibility is also a path we can go down if we expand usage of this model to include all of our logic

bkeepers:
> Are you suggesting that this PR should also refactor all of our unfurl logic so that it moves to the model? I think I'm cool with that, although it'll probably amount to a rather sizeable PR?

Yeah, I think it would make sense to do it all at the same time since refactoring the existing code will likely influence how the data gets persisted.

wilhelmklopp:
This was largely included in #498

PR #418, opened 2018-03-03
Update on issues.edited

wilhelmklopp:
We don't currently update the issue message when we receive the `issues.edited` event. Somehow I thought we already did that?

Either way, now we do :D


lib/activity/issues.js
```diff
@@ -21,7 +21,7 @@ async function matchMetaDataStatetoIssueMessage(context, subscription, slack) {
   const cacheKey = subscription.cacheKey(`issue#${context.payload.issue.id}`);
   const storedMetaData = await cache.get(cacheKey);
   if (storedMetaData) {
-    updateIssueMessage(storedMetaData, context, subscription, slack);
+    await updateIssueMessage(storedMetaData, context, subscription, slack);
```
wilhelmklopp: missing awaits for days

PR #423, opened 2018-03-05
Avoid linking to page that 404s

### LINKED ISSUE #316: 404 when clicking link to install GitHub App

bkeepers:
GitHub Apps currently return a 404 when linking directly to the new installation page for an organization that you don't have owner permissions on. This updates the link to link to the new installation page, where a user can then select the organization that they have permissions on.

Fixes #316 

Note that this will likely lead to more reports of #389. I expect users won't have permission to install the app on the organization, so they'll install it on their own user account, which is always the first in the list, and then get redirected back to the new installation page when we try to run the subscribe command again.

### DIFF
--- a/lib/github/index.js
+++ b/lib/github/index.js
@@ -22,7 +22,7 @@ module.exports = (robot) => {
     const info = await robot.info();
 
     req.log({ trigger_id, owner_id: req.params.id }, 'Redirecting to install GitHub APP');
-    res.redirect(`${info.html_url}/installations/new/permissions?target_id=${req.params.id}`);
+    res.redirect(`${info.html_url}/installations/new`);
   });
 
   const slackAppUrl = `https://slack.com/app_redirect?app=${process.env.SLACK_APP_ID}`;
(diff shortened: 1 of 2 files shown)


wilhelmklopp:
> Note that this will likely lead to more reports of #389. I expect users won't have permission to install the app on the organization, so they'll install it on their own user account, which is always the first in the list, and then get redirected back to the new installation page when we try to run the subscribe command again.

In addition to redirecting them after we validate that they still don't have the GitHub App installed, could we also `chat.postEphemeral` a message to them in the right channel which says something like 'you've installed the app on org x, but in order to subscribe to repo y/z you need to install it on org y'

PR #429, opened 2018-03-06
Handle repository deletions

### LINKED ISSUE #301: Disable subscriptions when a repository is deleted
Possibly the cause for #300

wilhelmklopp:
At the moment when a repository gets deleted we don't do anything about it.

With this PR, when a repository gets deleted, we post a message to the relevant channels (example below) and delete the subscription:
image

Closes #301


lib/activity/router.js
```diff
@@ -27,6 +27,13 @@ module.exports = ({ models }) => {
             include: [GitHubUser],
           });
 
+          const eventType = `${context.event}.${context.payload.action}`;
+
+          // Can't check for repo access because repo has been deleted
+          if (eventType === 'repository.deleted') {
+            return callback(context, subscription, slack);
+          }
+
```
wilhelmklopp: @bkeepers I wonder if there's a better way to do this

lib/models/subscription.js
```diff
@@ -24,6 +24,10 @@ module.exports = (sequelize, DataTypes) => {
     reviews: false,
   };
 
+  const FixedSettings = {
```
wilhelmklopp: This should probably have a better name - and maybe not even include the word 'settings'?

bkeepers: Hmm, should this be the same setting as `public` events? It looks like they're included in the same `repository` webhook: https://developer.github.com/v3/activity/events/types/#repositoryevent.

wilhelmklopp: the `public` event actually seems to come in via two webhooks. One as part of the repository event and one as part of the public event: https://developer.github.com/v3/activity/events/types/#publicevent

Right now we listen to the public event

lib/messages/repository-deleted.js
```diff
@@ -0,0 +1,26 @@
...
+
+  toJSON() {
+    const pretext = `Repository deleted by ${this.sender.login}`;
+
+    return {
+      attachments: [{
+        ...super.getBaseMessage(),
+        fallback: pretext,
+        pretext,
+        text: 'The subscription to this repository has been disabled.',
```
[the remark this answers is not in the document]
wilhelmklopp: Yeah, that's a really good point. 👍

PR #430, opened 2018-03-06
 Silence error logging done by @slack/client independently of try catch

### LINKED ISSUE #354: Slack error during migration:  `service_removed`
https://sentry.io/github-integrations/slack/issues/471109303/

```
Response not OK:  invalid_service
```

Appears to be coming from: https://github.com/integrations/slack/blob/cad3ad0b0e7d5ab3efa535d5da8c7591786e7c16/lib/models/legacy-subscription.js#L111

We've seen this error > 100 times in the last 24 hours.

cc @bertrandom @wilhelmklopp

wilhelmklopp:
The node-slack-sdk logs errors even when we catch them, which results in us still seeing the expected `service_removed` error in Sentry: https://sentry.io/github-integrations/slack/issues/471827621/

This PR sets the log level to 'fatal' for the duration during which we might get a `service_removed` error logged.

Ideally we'd achieve this by telling the node-slack-sdk to only throw errors the normal way and not also to log them, but not sure what the best way to achieve that is.

Closes #354


wilhelmklopp:
> 😕 This will actually have side-effects and hide logs for requests being handled while _makeAPICall is being processed.

ah, that's annoying :/ Let's close this PR then

> A few other ideas to solve this:

> 1. Just mark the errors as ignored in Sentry
> 1. Modify the [`logger` passed to the slack client](https://github.com/integrations/slack/blob/master/lib/slack/client.js#L7-L10) to avoid logging certain errors
> 1. See if there's a way we can pass a custom logger to `_makeAPICall`

Which out of these 3 would be your preferred solution? At the moment I'm kind of leaning towards just ignoring the error in Sentry

bkeepers:
> Which out of these 3 would be your preferred solution? At the moment I'm kind of leaning towards just ignoring the error in Sentry

Works for me!

PR #433, opened 2018-03-07
WIP: command interface

bkeepers:
The way we handle commands has not aged well. Commands are hard to test without doing full integration tests, and they're brittle and hard to follow after #311 and #371. I'd like to revisit this and in the process begin extracting some of the Slack bits into an adapter for Probot.

This pull request is pseudo code experimenting with what an ideal command interface might look like.

### Requirements

- Commands should be easy to define, grok, and test
- Command should be able to be invoked directly by Slack, invoked by other commands, or saved and invoked later by `trigger_id` once preconditions are met.
- The command logic shouldn't have to care about how it's invoked.

### Proposed API

`robot.command` defines new commands. A command definition is an object that has at least a `name`, `description`, and `action`, and optionally a `usage` hint.

```js
robot.command({
  name: 'echo',
  usage: 'echo text to echo'
  description: 'Echo whatever I say back to you.'
  action: command => {
    return command.respond(command.text);
  }
});
```

Under the hood, `robot.command()` is just shorthand for `robot.command.define`. `robot.command` also exposes these methods:

- `use(fn)` - register middleware that are run in order whenever a command is invoked. `fn` will receive the same `command` argument that the `action` of the command definition receives, and a `next` function to pass control to the next middleware.
- `define({name, description, action})` - define a new command. This is really just syntactic sugar for adding a middleware that checks if the `name` of the defined command matches the invoked command and calling `next` or `action`.
- `invoke({name, respond, text, ...args})` - invokes a command by name, a `respond` function that the command should call when it is done.
- `handler(req, res, next)` - an express middleware handler that will be used to receive commands from Slack and invoke the commands.

The `command` argument has these properties:

- `text` - the input of text for the command
- `respond(message)` - sends a response to the command
- All other [properties coming from Slack](https://api.slack.com/slash-commands#how_do_commands_work)

For anything not defined here, `robot.commands` will look to [koa](http://koajs.com/) for inspiration on middleware and error handling, and [commander.js](https://github.com/tj/commander.js) for inspiration on implementation.

### Prior art:

- https://github.com/hubotio/evolution/pull/2 and [hubot-commands](https://github.com/bkeepers/hubot-commands)
- [probot-commands](https://github.com/probot/commands)

I'd love feedback on this: cc @tcbyrd @JasonEtco @gr2m @technicalpickles @philschatz


JasonEtco:
I dig the API for it so far (especially `command.respond`).

Do you see this as being something specific to the Slack adapter, or would it be tacked on to the Robot class, then resolve the same API for the other services? AKA would you be able to type `/do-something` in a GitHub comment and still use this API, or is this Slack-only?

bkeepers:
I see this replacing https://github.com/probot/commands and being available for all adapters. The code currently in probot/commands will be moved into the GitHub adapter (since GitHub doesn't inherently support commands). Adapters will call `robot.commands.invoke(…)` with whatever command it receives.

There's still some work to do to standardize the invocation arguments.

philschatz:
I like the idea, especially that `command.respond(message)` would either respond with a Slack message or a GitHub Issue Comment depending on how the command was invoked.

It would be nice to define whether the commands support slack, GitHub, or both.

And in the case of Slack commands, would there be a nice way to support [Interactive Messages](https://api.slack.com/interactive-messages)?

Maybe something like the following which would implement the [Slack coffeebot example](https://github.com/slackapi/sample-message-menus-node):

```js
robot.command({
  name: 'coffee',
  usage: 'coffee',
  description: 'Order a coffee.',
  supports: ['SLACK'] // register only with Slack
  action: async (command, previousMessageToken) => {
    const state = command.getState()
    if (!state) {
      return command.respond('I am coffeebot and I am here to take your order. What would you like to do?', {
        attachments: [
          { 
            type: 'button',
            color: '#5A352D',
            text: 'Start a coffee order',
            onSelect: async () => command.extendState({orderStep: 'LOAD_OPTIONS'})
          }
        ]
      });

    } else if (state['orderStep'] === 'LOAD_OPTIONS') {
      // This edits the current message instead of creating a new one
      return command.edit(previousMessageToken, 'Great! What can I get started for you?', {
        attachments: [
          {
            type: 'select',
            text: 'Select the type of coffee you would like to order',
            options: [
              { text: 'Cappuccino', onSelect: () => { comamnd.extendState({orderStep: 'CHOOSE_MILK', coffeeType: 'cappuccino'}) } },
              { text: 'Americano',  onSelect: () => { comamnd.extendState({orderStep: 'CHOOSE_STRENGTH', coffeeType: 'americano'}) } },
            ]
          }
        ]
      })
    } else if (state['orderStep'] === 'CHOOSE_MILK') {
      // ...
    } else {
      // clear the state upon sending this message because we are done.
      return command.finish(previousMessageToken, `Your ${state['coffeeType']} is ready!`)
    }
  }
});
```

`command` could have the following additional methods:

- `getState()` : returns the current state object
- `extendState({key: value})` : returns a `Promise`
- `edit(messageToken, text, slackOptions)` : Replaces the contents of an existing message
- `finish(optionalMessageToken, text, slackOptions)` : Completes the workflow and discards the state


And every time `comamnd.extendState(...)` is called, `action(...)` would be invoked.

🤔 There might be an even-simpler way to write this example...

wilhelmklopp:
Very excited for this!

Immediate first thought: I wonder if we should further separate the command (input) and the action (output). Slash commands are one way for a user to take an action, but especially in Slack so are buttons.

For buttons, would there be a `robot.button()` which perhaps invoke the same action as the command?

As an example for the GitHub Slack integration, in the message in which we prompt users to re-activate their legacy subscriptions, we currently say something along the lines of "Run `/github subscribe owner/repo` to re-activate." An improved way to achieve the same thing would be to present the user with a button that just says "Re-activate owner/repo". 

From our app's perspective that should invoke the same "action" that the slash command logic would ultimately invoke, but we see the incoming event from the button differently that we would from the slash command.

ntwb:
> An improved way to achieve the same thing would be to present the user with a button that just says "Re-activate owner/repo".

That would then lead toward supporting _Interactive Messages_ as mentioned by @philschatz 🎉

philschatz:
Slack's Interactive Messages format (`{attachments: [ {type: 'button', ...`) might also be able to be used to create _Interactive_ GitHub commands. Slack has an option to add a Button or a Dropdown, and can either create a new message or edit an existing one.

In GitHub, both of those could be achieved by providing a link with a URL to the bot (for a `button`), or a list of links with URLs to the bot (for `select`). 

For operations that require the user to be authenticated, the bot could also be an OAuth client.

philschatz:
To test out the feasibility of adding Interactive Messages to Probot (in https://github.com/integrations/slack/pull/433#issuecomment-371392146 ) I created a simple Probot app using the pseudocode above. You can try it out here: https://github.com/philschatz/barista-bot/issues/1 or watch the screencap below:

### Screencap

[barista-bot](https://github.com/philschatz/barista-bot/issues/1)

It addresses the issue of how the request came in (from a `/command` or from a Button press) by whether the current state is empty (`/command`) or not (Button) but the same asynchronous `action` function is called in both cases.

JasonEtco:
Looking at this again after a little bit - where would the `description` field come up? I think its super valuable from a transparency point of view, but am I correct in assuming that it'd be invisible for all but Slack commands? In that case, would it be optional?

bkeepers:
Closing since this is out of date and there are not plans to move it forward.

Issue #315, opened 2018-02-19
Error when user revokes GitHub app authorization 


bkeepers:
**Steps to reproduce:**
1. Go to [Authorized GitHub Apps](https://github.com/settings/apps/authorizations) in your GitHub account settings
2. Click **Revoke** on the Slack app
3. Type `/github subscribe my/repo` in a slack channel
4. Observe error: "Could not find resource: `integrations/test`"

**Expected behavior:**
Prompt to sign in to GitHub again

It would be nice if GitHub Apps delivered an event to notify us that a user has revoked access to the app.

PR #444, opened 2018-03-08
More logging for failed access checks

bkeepers:
We're experiencing a lot of `401 Bad Credentials` errors from GitHub when performing this access check. This adds a little more logging so we can see who exactly this is affecting and dig into what might be causing it, and renames the error param to `err`, which is given special treatment by the the bunyan logger and thus sentry.

cc #388

### DIFF
--- a/lib/models/githubuser.js
+++ b/lib/models/githubuser.js
@@ -33,7 +33,7 @@ module.exports = (sequelize, DataTypes) => {
           if (err.code === 404) {
             return false;
           }
-          logger.warn({ error: err }, 'GitHub returned neither 200 nor 404');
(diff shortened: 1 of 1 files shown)


bkeepers:
@wilhelmklopp after testing this in production, I'm 99% certain that the 401s are coming from users who have revoked access from the GitHub App. I think the correct behavior is to treat a 401 as a 404 and disable the subscription.

We could do a custom error message for 401 that says something about the user revoking access.

cc #315

wilhelmklopp:
> We could do a custom error message for 401 that says something about the user revoking access.

Yeah, that seems like a good way of handling this. Should we open a separate issue for that?

PR #448, opened 2018-03-13
Update @keyv/redis to fix Redis out of memory issue

bkeepers:
We've twice now run into issues with Redis hitting an OOM error:

```
ReplyError: OOM command not allowed when used memory > 'maxmemory'.
```

The cause appears to be `@keyv/redis` using a `namespace:keyv` set to store every key that has ever been written to Redis, which is used to get a list of keys when for `cache.clear()`. I opened https://github.com/lukechilds/keyv-redis/pull/15 to change the behavior to use the built-in `LIST keyv:*` command.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,9 +36,7 @@
       }
     },
     "@keyv/redis": {
-      "version": "1.3.8",
-      "resolved": "https://registry.npmjs.org/@keyv/redis/-/redis-1.3.8.tgz",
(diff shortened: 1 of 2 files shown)


lukechilds:
Depending on how you're using Keyv and what other data you're storing in Redis, this may not actually be a very good solution and could lead to serious Redis performance degradation.

More info in the related issue:

https://github.com/lukechilds/keyv-redis/pull/15

PR #450, opened 2018-03-13
Document when links won't unfurl

bkeepers:
Since this has come up a few times via support, wanted to add something that we could link to in response to users.

cc @izuzak @francisfuzz @wilhelmklopp


bkeepers:
@izuzak 👍 thanks!

Issue #192, opened 2017-12-19
Timeouts in production


bkeepers:
From the Heroku dashboard:

> ⚠️  In the last 24 hours, there have been 1601 critical errors for this app.

Response time is 6ms for the 50th percentile, but 30s for the 95th percentile.

![](https://user-images.githubusercontent.com/173/34172336-4e5d5916-e4b7-11e7-81c7-1361bb6d5933.png)

The critical errors appear to be timeouts. My best guess is that they're coming from unhandled exceptions (like #154). We should also add some tests to ensure that runtime exceptions return a 500 and don't just leave the connection hanging until Heroku closes it.

bkeepers:
These are back. It looks like they started around #371. PR coming to add logging for request timeouts.

bkeepers:
Appears to be fixed by #452

PR #452, opened 2018-03-13
Redirect back to slack channel if resource can't be found

bkeepers:
It looks like many of the timeouts in production (#192) are coming from this issue.

Two ways to trigger this:

1. Type `/github subscribe owner/repo` with an `owner` that exists but does not have the GitHub App installed, and an `repo` that does not exist
1. User is prompted to install GitHub App
1. User clicks link, is taken to GitHub, installs the GitHub app
1. User is redirected back to integration, which fails to find the repo, but doesn't redirect the user back to slack so request times out and an error is posted into Slack

or…

1. Type `/github subscribe owner/repo` with either an org that doesn't have the app installed, or a the installation doesn't have access to `repo`
1. User is prompted to install GitHub App
1. User clicks link, is taken to GitHub, and installs the GitHub app, but doesn't give it access to `repo`
1. User is redirected back to integration, which fails to find the repo, but doesn't redirect the user back to slack so request times out and an error is posted into Slack

cc #311


lib/commands/subscribe.js
```diff
@@ -2,6 +2,8 @@ const {
   Subscribed, NotFound, AlreadySubscribed, NotSubscribed, UpdatedSettings,
 } = require('../messages/flow');
 
+const isSlackbot = require('../is-slackbot');
```
wilhelmklopp: nit: `isSlackbot` might confuse someone who's looking at this code in the future. Could we maybe call this something like `originatesFromSlack` or `isRequestFromSlack`?

PR #455, opened 2018-03-14
Always use command.respond

bkeepers:
#452 fixed most of the remaining timeouts in production, but there are still a few happening, and it's anywhere we're calling `command.respond` without handling the situation where a command might be called from a browser instead of from slackbot (#311).  These are only happening in edge cases where someone types the command wrong (`/github subscribe nothing`).

This removes the hacky `respondWith` that was used in the `subscribe` command, and instead adds logic to redirect back to the slack channel whenever calling `command.respond` from within a web request.

The only time we'll still need special handling for a request from the browser now is if you want to send the user somewhere besides the slack channel that they came from (e.g.  OAuth or app installation URL).

This is an incremental improvement as we work towards #433

### DIFF
--- a/lib/commands/subscribe.js
+++ b/lib/commands/subscribe.js
@@ -2,8 +2,6 @@ const {
   Subscribed, NotFound, AlreadySubscribed, NotSubscribed, UpdatedSettings,
 } = require('../messages/flow');
 
-const isRequestFromSlack = require('../is-request-from-slack');
-
 /**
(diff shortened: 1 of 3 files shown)


bkeepers:
This has been in production for a couple hours now with no issues, and it did resolve the remaining timeout issues we were having. 

@wilhelmklopp review is still welcome on this and I can submit a followup PR to fix any issues, but I want to get this out since it fixes a user-facing issue.

PR #470, opened 2018-03-16
Include PR's author in review messages

devrelm:
This PR adds the PR author's username to the `pull_request_review` messages. This way, the PR author can be notified when someone approves, rejects, or comments their PRs.

#### Use Case

My team uses this integration, and we have each added our github username to the list of Slack keywords that we're notified about. Unfortunately, this hasn't actually been too useful since we're only notified for the actions that we ourselves took (like opening a PR.)


bkeepers:
Thanks @devrelm! This is being deployed now.

devrelm:
Awesome, thanks! I just noticed it when someone did an early morning review, and already let the team know to add their GH username as a slack notification keyword if they hadn't already. Cheers!

PR #481, opened 2018-03-20
Render comments and reviews

### LINKED ISSUE #461: Markdown in comments does not get rendered
As originally posted at https://github.com/integrations/slack/issues/408#issuecomment-373049194, Markdown syntax in comments does not get rendered on the Slack notification side as expected:

From https://bisq.slack.com/archives/C7UHH8UQM/p1521038700000431:

image

bkeepers:
Fixes #461 by fetching `body_html` for comments.


lib/activity/comments.js
```diff
@@ -2,10 +2,24 @@ const cache = require('../cache');
...
+  // Fetch updated comment to get body_html
+  try {
+    comment = (await context.github.request({
+      method: 'GET',
+      url: context.payload.comment.url,
+      headers: { accept: 'application/vnd.github.html+json' },
+    })).data;
+  } catch (err) {
+    // Review comments API is not yet accessible by GitHub Apps
+    context.log.warn({ err }, 'Could not fetch HTML for comment');
```
bkeepers: This is causing a _ton_ of exceptions in production. I'm re-deployed master and will see if we can get GitHub App support for `/repos/:owner/:repo/pulls/comments/:id`

lib/activity/comments.js
```diff
@@ -2,10 +2,26 @@ const cache = require('../cache');
...
+    comment = (await context.github.request({
+      method: 'GET',
+      url: context.payload.comment.url,
+      headers: { accept: 'application/vnd.github.html+json' },
+    })).data;
+  } catch (err) {
+    // Review comments API is not yet accessible by GitHub Apps
+    if (err.code !== 403) {
+      throw err;
+    }
```
bkeepers: @wilhelmklopp what do you think about this? The review comments API is accessible by GitHub Apps yet, but everything else is. This allows mrkdwn rendering to work for all other comment types while we wait for that API to be enabled, and then we can remove this catch once it's done.

wilhelmklopp: Works for me! Could we maybe also include a logger.warn above the `throw err` that says something like 'Review comments not accessible to GitHub Apps', so that we can differentiate this error from all the other 403s we might get?

bkeepers: Just to clarify, this will `catch` 403s and not report them. So the only other unexpected errors will get reported.

I had a `warn` before 5b2752c, but it causes a ton of unnecessary warnings to get reported in Sentry.

PR #486, opened 2018-03-21
Add repository name to push notification fallback text

### LINKED ISSUE #485: Slack fallback text doesn't include repo information
The slack app uses the `fallback` property of the received message as the text of any desktop/mobile notifications it triggers.

I appreciate the intention to keep the notification text brief, but without the repo information I find I get a series of `1 commit pushed to master` notifications without the context of which repo it's related to.

In our project we have a single #github channel subscribed to all the project's repos. Some I care about more than others.

Having just upgraded from the old github slack app, I find I can no longer mentally filter the notifications I care about at a glance - I have to go look at slack to get that small bit of extra detail.

Would you consider a PR that adds the repo name to the name of the notification text?

I think the simple change would be to add the appropriate bit just to the `fallback` property of the slack message: https://github.com/integrations/slack/blob/master/lib/messages/push.js#L38

knolleary:
Fixes #485 

Have not added any explicit new tests as it appears well covered by existing tests once they were updated to expect the additional text.


lib/messages/push.js
```diff
@@ -35,7 +35,7 @@ class Push extends Message {
       attachments: [
         {
           ...super.getBaseMessage(),
-          fallback: this.summary,
+          fallback: `${this.summary} - ${this.push.repository.full_name}`,
```
bkeepers: I think the old integration did `[owner/repo] - summary`. I don't have a strong preference either way. The only reason I could see to put repo name first is that it's typically a shorter and more predictable length than other parts of the message (maybe not specifically for push, but probably for other messages).

knolleary:
You're quite right - it would be better to prefix the repo.

And as I think of it, this should be applied across all the different message types - not just Pushes as this PR focusses on.

I'm happy to go through the other message types and add it in for consistency as part of this PR. Day job in the way right now - will get it updated tonight.

bkeepers:
👍 Sounds great. Thanks!

knolleary:
Have now updated this PR to prefix the message with the repo name, and applied it to a wider range of messages.

lib/messages/blob.js
```diff
@@ -39,7 +40,7 @@ class Blob extends Message {
 
     const tripleBackTick = '```';
     return {
-      fallback: title,
+      fallback: `[${this.repository.full_name}] - ${title}`,
```
wilhelmklopp: We don't ever post an activity message for blobs, so would this ever actually show up in a notification? I'm not sure how the fallback works for unfurls, but I have a feeling the Slack notification is just the link that the user posted.

knolleary:
I think I've updated all the remaining notifications.

I have *not* changed the `public` one - as it already starts with the repo name, it seemed excessive to add it again.

wilhelmklopp:
@knolleary 🤘 I think this is almost ready to go out.

> I have not changed the public one - as it already starts with the repo name, it seemed excessive to add it again.

That makes sense 👍 

Did you have any thoughts on my comment above? https://github.com/integrations/slack/pull/486#discussion_r176956272

Also just to get an idea of what this change will actually look like: Could you post example screenshots of a desktop notification and a mobile notification?

I'm wondering if we can even drop the `-` and just have it be `[owner/repo] Pulll request opened by knolleary`

knolleary:
RE blob unfurls - I'm not sure what the right answer is either. I was going for consistency here - if the fallback is never used, then I figured no harm in updating them.

I agree removing the `-` is a good idea - ensure more of the real message is seen. I can go through and update them.

As for screenshots, I haven't got a fully fledged copy of this integration setup locally - this was 'only' changing the text of the notifications that the unit tests could verify. I'll see if I can get a [copy running](https://github.com/integrations/slack/blob/master/CONTRIBUTING.md#configuring-a-slack-app) but it may be a while before I have the time to do that properly.

wilhelmklopp:
> As for screenshots, I haven't got a fully fledged copy of this integration setup locally - this was 'only' changing the text of the notifications that the unit tests could verify. I'll see if I can get a copy running but it may be a while before I have the time to do that properly.

Yeah, there are quite a few steps in getting it running locally. I can grab those screenshots :)

bkeepers:
Tested on mobile and it looks good (well…the changes here do anyway. The `#general @github (bot):` part seems noisier than it should be)

img_0977

knolleary:
Looks good 👍

bkeepers:
Thanks @knolleary!

PR #489, opened 2018-03-23
Rename error-hander.test.js to error-handler.test.js

JasonEtco:
Does what the title says ☝️ 

There is a file name with a typo in it, `error-hander` should be `error-handler`.

🌮


bkeepers:
👍 thanks

Issue #335, opened 2018-02-22
Rollups not working for PRs


jchuerva:
*Scenario*
* Create a PR -> Notification in slack is displayed
* Update the PR adding a label

*Expected*
* Previous PR slack notification is updated with the new label

*Actual*
* Previous PR slack notification is not updated

PR #494, opened 2018-03-27
Mitigate replication lag for issues and PRs

wilhelmklopp:
- Implements some logic that delays some GitHub API requests so that we're less likely to run into problems relating to replication lag
- Removes the API requests completely for issue and pr events where we don't need to make a separate API call. (All relevant data is included in the webhook)

### DIFF
--- a/lib/activity/pull-requests.js
+++ b/lib/activity/pull-requests.js
@@ -1,18 +1,12 @@
-const { promisify } = require('util');
-
 const { PullRequest } = require('../messages/pull-request');
 
 const cache = require('../cache');
-
-const sleep = promisify(setTimeout);
 
 async function pullRequestEvent(context, subscription, slack) {
   let pullRequest = {
     ...context.payload.pull_request,
   };
   if (context.payload.action === 'opened') {
-    await sleep(1000);
-
     // need seperate api request to get labels and body_html
     const issue = (await context.github.issues.get(context.issue({
       headers: { accept: 'application/vnd.github.html+json' },
--- a/lib/activity/pull-requests.js
+++ b/lib/activity/pull-requests.js
@@ -1,19 +1,31 @@
+const { promisify } = require('util');
+
 const { PullRequest } = require('../messages/pull-request');
 
 const cache = require('../cache');
 
-async function pullRequestEvent(context, subscription, slack) {
-  const eventType = `${context.event}.${context.payload.action}`;
-  // need seperate api request to get labels and body_html
-  const issue = (await context.github.issues.get(context.issue({
-    headers: { accept: 'application/vnd.github.html+json' },
-  }))).data;
+const sleep = promisify(setTimeout);
 
-  const pullRequest = {
+async function pullRequestEvent(context, subscription, slack) {
+  let pullRequest = {
     ...context.payload.pull_request,
-    labels: issue.labels,
-    body_html: issue.body_html,
   };
+  if (context.payload.action === 'opened') {
+    await sleep(1000);
+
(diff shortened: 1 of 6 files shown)


bkeepers:
Alternate suggestion: update `router.js` and register an octokit plugin to sleep for a second the first time any read happens.

```js
const { promisify } = require('util');
const sleep = promisify(setTimeout);

function avoidReplicationLag() {
  let shouldSleep = true;

  return async (options) => {
    // Only GET requests are affected by replication lag 
    if(shouldSleep && options.method === 'GET') {
      await sleep(1000);
    }
    shouldSleep = false;
  }
}

// in router.js

octokit.hook.before('request', avoidReplicationLag());
```

wilhelmklopp:
> Alternate suggestion: update router.js and register an octokit plugin to sleep for a second the first time any read happens.

Just thinking about if there are any cases of activity messages where it doesn't make sense to delay GET requests, but I think there aren't.

The only other case besides issues and PRs where we even make a GET request is for statuses, and we might very well run into the same replication lag issue there although due to how we've implemented it at the moment we haven't run into that problem a lot.

lib/activity/router.js
```diff
@@ -55,6 +58,9 @@ module.exports = ({ models }) => {
             return;
           }
 
+          // Delay GitHub API calls to avoid replication lag
+          context.github.hook.before('request', avoidReplicationLag());
+
```
bkeepers: Any way to remove this duplication? Maybe switch the conditional above to be `if (!subscription.creatorId)`, do the permission check, and then only call `callback` once at the end?

lib/activity/pull-requests.js
```diff
@@ -3,17 +3,23 @@ const { PullRequest } = require('../messages/pull-request');
...
+  let pullRequest = {
     ...context.payload.pull_request,
-    labels: issue.labels,
-    body_html: issue.body_html,
   };
+  if (context.payload.action === 'opened') {
+    // need seperate api request to get labels and body_html
+    const issue = (await context.github.issues.get(context.issue({
+      headers: { accept: 'application/vnd.github.html+json' },
+    }))).data;
```
bkeepers: Why only `opened`? What happens if I edit a PR?

lib/activity/router.js
wilhelmklopp: I like that 👍

lib/activity/pull-requests.js
wilhelmklopp: We don't actually do anything at the moment when you edit a PR. That will be implemented in https://github.com/integrations/slack/issues/335

We only route the following to this function:
https://github.com/integrations/slack/blob/master/lib/activity/index.js#L36

bkeepers:
Finally tracked down the production issue and it's being caused by `newrelic`. This code works fine on node 8.0+ without the newrelic agent:

``` 
const sleep = require('util').promisify(setTimeout)

sleep(1000).then(() => console.log('done sleeping')) 
```

However, it fails whenever the newrelic agent is loaded with:

``` 
TypeError: "callback" argument must be a function 
File "timers.js", line 434, col 11, in setTimeout 
File "/app/node_modules/newrelic/lib/transaction/tracer/index.js", line 337, col 23, in wrappedFunction 
return original.apply(this, arguments) 
File "internal/util.js", line 230, col 26, in wrappedFunction 
```

wilhelmklopp:
One thing that would have definitely sped finding this up is if we had a proper stack trace for the error: https://sentry.io/github-integrations/slack/issues/505692768/

That's perhaps something else worth mentioning to newrelic support and also to Sentry.

PR #498, opened 2018-03-29
WIP: Private unfurls team ship

wilhelmklopp:
Private unfurls as a whole changes quite a few things of how unfurls currently work and doing them all at once has proved challenging. 

This PR will implement a very minimal first step towards private unfurls. Private unfurls as a feature will be limited to only a small amount of slack channels (2) within 2 separate workspaces.

This first step towards private unfurls is the following:
- All unfurls in the 2 team-ship channels will be using the user's access token, which means users need to have their GitHub account connected in order for any links they post to unfurl
- When the link is private, the user will get a one-off prompt in the form of a button to unfurl. 

Todo: 
- [ ] remove code that won't be included in this ship
- [ ] implement the team-ship functionality and make sure old functionality doesn't break
- [ ] implement 'no' option for prompt
- [ ] Add and modify tests


wilhelmklopp:
Wow, codecov commenting **twice**. To what do I owe this honour?

lib/models/unfurl.js
```diff
@@ -0,0 +1,226 @@
...
+      }
+      // for now: if it's not part of a repo, then it can't be private
+      if (Object.keys(params).indexOf('repo') === -1) {
+        return false;
+      }
+      const repo = await github.repos.get({ owner: params.owner, repo: params.repo });
+      return repo.data.private;
+    },
+
+    async spawn({
```
wilhelmklopp: Not super happy with this name, but couldn't come up with anything better. Any ideas?

lib/unfurls/index.js
```diff
@@ -1,105 +1,88 @@
...
-      where: { slackId: req.body.team_id },
-    });
-
-    req.log.debug(unfurls, 'Unfurling links');
-    const slackRes = await workspace.client.chat.unfurl({
-      ts: event.message_ts,
-      channel: event.channel,
-      unfurls,
+async function unfurlAction(req, res) {
+  if (req.body.actions[0].name === 'unfurl-dismiss') {
```
wilhelmklopp: Question: When the unfurl is dismissed, should we delete the db record? Or should we have another `isDismissed` column that we set to `true` when it gets dismissed?

lib/middleware/route-action.js
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Routes the request based on event type
+ */
+module.exports = function route(req, res, next) {
+  const action = req.body.callback_id.substr(0, req.body.callback_id.indexOf('-'));
+
+  if (action) {
+    logger.debug({ ...req.body, token: undefined }, 'Action received from Slack');
+    // Append action derived from callback_id to the request URL to route accordingly
+    req.url += `:${action}`;
```
wilhelmklopp: Chose a `:` here to differentiate it from the way `route-events` works. Should we stick with that or use a `.` here just like for `route-events`

lib/models/unfurl.js
```diff
@@ -0,0 +1,226 @@
...
+      const params = githubUrl(url);
+      if (!params || !resources[params.type]) {
+        throw new UnsupportedResource(url);
+      }
+      const attachment = await resources[params.type](params, github, unfurlType);
+
+      // Add lock to the footer if resource is not public
+      if (!isPublic) {
+        attachment.footer = `:lock: ${attachment.footer}`;
+      }
```
wilhelmklopp: Here is what this looks like. Feels kinda weird to do this outside of the renderer/messages folder but implementation wise this solution has far fewer lines of code.

image

lib/models/unfurl.js
```diff
@@ -0,0 +1,226 @@
...
+      this.isDelivered = true;
+      await this.save();
+    },
+  });
+
+  Object.assign(Unfurl, {
+    async getGitHubClient(channelId, userSlackId) {
+      const github = new GitHubApi({ logger });
+
+      if (!process.env.ALLOWED_CHANNELS || !process.env.ALLOWED_CHANNELS.split(',').includes(channelId)) {
```
bkeepers: > `process.env.ALLOWED_CHANNELS.split(',')`

This would be good to do once at the top of the file and assign to a constant to avoid having to split the string on every request.

lib/models/unfurl.js
```diff
@@ -0,0 +1,226 @@
...
+          });
+          logger.debug({
+            channelId,
+          }, 'using PAT authed github client for unfurl');
+        }
+        return github;
+      }
+
+      const slackUser = await SlackUser.findOne({
+        where: { slackId: userSlackId },
```
bkeepers: Does this also need to scope based on workspace?

lib/models/unfurl.js
```diff
@@ -0,0 +1,226 @@
...
+      };
+    },
+
+    async isPrivate(github, url) {
+      const params = githubUrl(url);
+      if (!params || !resources[params.type]) {
+        throw new UnsupportedResource(url);
+      }
+      // for now: if it's not part of a repo, then it can't be private
+      if (Object.keys(params).indexOf('repo') === -1) {
```
bkeepers: Try the [in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) operator:

```js
if (!'repo' in params) {
```

test/fixtures/index.js
```diff
@@ -37,6 +37,38 @@ module.exports = {
...
+          name: 'test',
+        },
+        user: {
+          id: 'U88HS',
+          name: 'aaron',
+        },
+        response_url: 'https://hooks.slack.com/actions/1234/5678',
+        trigger_id: '13345224609.738474920.8088930838d88f008e0',
+      }),
+    },
```
bkeepers: I know there are already a few things defined here (and they've annoyed me a few times), but how about defining this in `fixtures/slack/action.js` and requiring it here?

lib/models/unfurl.js
bkeepers: Heh, I almost commented about that. I loled at `spawn`. `prompt`?

lib/unfurls/index.js
bkeepers: Is there ay reason to keep it around? (Besides analytics, which we discussed using a different tool for those purposes)

lib/middleware/route-action.js
bkeepers: I don't feel strongly either way, and it's something we can easily change later. I'd like to get rid of this pattern of modifying the routes and go back to an event emitter eventually, but it's not urgent.

lib/unfurls/index.js
wilhelmklopp: No reason to keep it around besides analytics and building up an understanding of in which cases folks decide not to unfurl. Although this current implementation is doesn't fully serve the analytics use case. To fully serve that use case we'd need `isDismissed`, which is probably not a good thing to implement at this stage. 

I'll add some code to implement the deleting logic.

lib/models/unfurl.js
wilhelmklopp: Just tried implementing this. It works fine in `unfurls/index.js`, but not in `models/unfurl.js` because that file is only required once. Probably fine for staging/production but it breaks testing, so I'm just going to stick to the existing implementation for now.

lib/models/unfurl.js
wilhelmklopp: Yep 👍 Changing this to use our internal id 😬

lib/models/unfurl.js
wilhelmklopp: ooo I like prompt! I think i'm going with `promptOrDeliver`

PR #503, opened 2018-03-29
Cache user lookup

### LINKED ISSUE #500: Cache `hasRepoAccess` for 10 minutes
We're starting to see the impact of not caching this function, which is an extremely hot code path. (2 pages today)

The problem has manifested itself in the form of rate limiting based on the user who set up the subscription. This has the potential for negative consequences. Currently those negative consequences should only be noticeable to users when setting up new subscriptions, but in the future this will impact unfurls as well.

https://github.com/integrations/slack/blob/92b52d6a144b8b4a2db93b0050a63bfbca4d7d51/lib/models/githubuser.js#L37-L47

bkeepers:
This adds `cache.fetch(key, callback)` as shorthand for fetching a value from the cache or setting it if it doesn't exist, and uses it to look up if the creator still has access to a repository.

Fixes #500


lib/activity/router.js
```diff
@@ -26,7 +27,12 @@ module.exports = function route(callback) {
             include: [GitHubUser],
           });
 
-          if (!await creator.GitHubUser.hasRepoAccess(subscription.githubId)) {
+          const hasRepoAccess = await cache.fetch(
+            subscription.cacheKey('creator-access'),
+            () => creator.GitHubUser.hasRepoAccess(subscription.githubId),
+          );
```
wilhelmklopp: What are your thoughts on having a TTL of `10 * 60 * 100` here (10min) given that the hasRepoAccess check is a security feature.

wilhelmklopp:
Thinking about this some more, I can imagine a scenario where this could get quite confusing for users.
1. User gets removed from a repo by mistake
2. After 10min this check happens and the user is determined not to have access anymore, so the subscription gets deleted and a prompt gets sent to Slack
3. The user is confused about why they lost access and pings their admin. The admin realized they made an error and reinstates the relevant permissions.
4. The user proceeds to reactive the subscriptions which works fine, but once another activity message delivery comes in, the subscription deactivates again because we've still cached `hasRepoAccess` as `false`, so for the next 10min things will look super confusing to the user.

The question that comes out of this: Should we only cache `hasRepoAccess` when it's `true` and not when it's `false`?

lib/activity/router.js
bkeepers: 👍 That's fine for me. `fetch` takes a ttl as a 3rd arg.

bkeepers:
Right now reactivating the subscription will create a new subscription, so it will be a new cache key.

If we ever get to just disabling subscriptions, we could delete the cache key when deactivating/reactivating the subscription.

lib/cache.js
```diff
@@ -8,4 +8,15 @@ const keyv = new Keyv({
 
 keyv.on('error', logger.error.bind(logger));
 
+async function fetch(key, callback, ttl = undefined) {
```
wilhelmklopp: I just realized: Isn't `ttl` going to be undefined here regardless of the ` = undefined`

```
> function a (x, y, z) { console.log(x, y, z)}
undefined
> a()
undefined undefined undefined
undefined
> a(1)
1 undefined undefined
undefined
```

bkeepers: Yeah, but this just makes it more clear to the reader that the 3rd argument is optional and is ok to be undefined.

PR #504, opened 2018-03-30
Update html-to-mrkdwn@2.0.0

bkeepers:
Improvements:

- Remove language in highlighted code block (https://github.com/integrations/html-to-mrkdwn/pull/14)
- Fix unnecessary escaping (https://github.com/integrations/html-to-mrkdwn/pull/15)
- Return the first image so it can be attached to rendered messages (https://github.com/integrations/html-to-mrkdwn/pull/16)


lib/messages/abstract-issue.js
```diff
@@ -70,7 +70,9 @@ class AbstractIssue extends Message {
     };
     if (this.major) {
       if (this.abstractIssue.body_html) {
-        core.text = mrkdwn(this.abstractIssue.body_html);
+        const { text, image } = mrkdwn(this.abstractIssue.body_html);
+        core.text = text;
+        core.image_url = image;
```
wilhelmklopp: do we need a conditional here in case there is no image? I suppose Slack is fine with us passing `image_url: undefined`?

lib/messages/comment.js
```diff
@@ -30,10 +27,18 @@ class Comment extends Message {
...
       author_icon: this.comment.user.avatar_url,
-      text,
       ts: moment(this.comment.created_at).unix(),
       mrkdwn_in: ['text'],
     };
+
+    if (this.comment.body_html) {
+      const { text, image } = mrkdwn(this.comment.body_html);
+      baseMessage.text = text;
+      baseMessage.image_url = image;
```
wilhelmklopp: same question here

bkeepers:
👍 added tests and updated snappydoo for the new tests

PR #511, opened 2018-04-05
WIP Private unfurl flow mockups

wilhelmklopp:
Before we can make private unfurls generally available we need implement two flows, which both require user settings: 
- [ ] Users should be able to automatically
   - [ ] unfurl links from a specific private repo (or org?)
   - [ ] unfurl links to all private resources they post

- [ ] Users should be able to silence prompts
  - [ ] for the next 24h
  - [ ] indefinitely

Letting users set settings as part of this flow brings up another piece of work that should be considered: Letting users alter the settings outside of the prompt->unfurl flow

- [ ] Users should be able to run `/github settings` to
  - [ ] adjust which private repos automatically get unfurled
  - [ ] un-silence unfurl prompts

cc @cmwinters for 💭

### DIFF
--- a/docs/private-unfurls.md
+++ b/docs/private-unfurls.md
@@ -1,0 +1,43 @@
+# Private unfurls
+
+Below are mockups for potential flows that need to be implemented before we can make private unfurls generally available.
+
+## Current flow
+
+Show Rich preview
+![Rich preview demo](https://d2mxuefqeaa7sj.cloudfront.net/s_4ACE98FACF4997A859FBDCF23940D470F9FBABA31346EEED9CA69C21A513E7A0_1522849637496_rich+preview.gif)
+
+Dismiss
+![Dismiss demo](https://d2mxuefqeaa7sj.cloudfront.net/s_4ACE98FACF4997A859FBDCF23940D470F9FBABA31346EEED9CA69C21A513E7A0_1522849647345_dismiss.gif)
+
+## Potential flow
+
+`1`. Initial prompt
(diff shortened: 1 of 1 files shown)


docs/private-unfurls.md
```diff
@@ -0,0 +1,43 @@
...
+
+Show Rich preview
+![Rich preview demo](https://d2mxuefqeaa7sj.cloudfront.net/s_4ACE98FACF4997A859FBDCF23940D470F9FBABA31346EEED9CA69C21A513E7A0_1522849637496_rich+preview.gif)
+
+Dismiss
+![Dismiss demo](https://d2mxuefqeaa7sj.cloudfront.net/s_4ACE98FACF4997A859FBDCF23940D470F9FBABA31346EEED9CA69C21A513E7A0_1522849647345_dismiss.gif)
+
+## Potential flow
+
+`1`. Initial prompt
```
wilhelmklopp: I feel like maybe this initial prompt should have a 'Don't ask me again' button in addition to 'show rich preview' and 'dismiss. That would then trigger the 'Getting too many of these prompts?' flow

bkeepers:
Just for a though exercise, I wonder what we could do to avoid prompting at all. Here are a few scenarios:

1. The channel is already subscribed to activity for that repository?  👍
1. Multiple links to a repository have been shared in the same channel recently? 🤔
1. Either of the above for another private repository in the organization? 🤔(seems a little risky, but just brainstorming)

Any others?

PR #521, opened 2018-04-10
Potential fix for smee

orta:
(Disclaimer, after enough time I just switched to ngrok (which I know well) so I didn't get all the way using smee, but spotted this.)

This is a bit more of a question PR, but I don't think smee is set up right for this project. 

Smee would get set up with a root like this:

screen shot 2018-04-09 at 5 16 19 pm

Which would work fine for a plain old probot - but this isn't a plain old probot, this repo needs other webhooks. So you have no way of going to the project root to do the GH oauth stuff, nor any way to hook up the slack calls. 

This PR moves the root back to the root.

screen shot 2018-04-09 at 5 34 26 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "start": "probot run --webhook-path=/github/events ./lib",
+    "start": "probot run ./lib",
     "test": "jest --coverage --runInBand --forceExit",
     "posttest": "npm run lint",
     "lint": "eslint lib test",


orta:
OK, yeah, this breaks my ngrok setup - to make this work, the PR would need to tell probot to use `github/events` but to not tell smee to set that as the root.

bkeepers:
Sorry for the confusion. Smee was changed to the default for Probot after these docs were written. You'll still need to use localtunnel or ngrok for local development with this app.

Let's get a PR open documenting that.

PR #525, opened 2018-04-13
Respond to `link_shared` event immediately

wilhelmklopp:
We currently do all of our unfurl processing before we respond to the `link_shared` event that we receive from Slack. This can sometimes take longer than 3s, which causes Slack to re-send the `link_shared` event leading to a number of race conditions.

This is only rarely visible to users, but it does lead to duplicate unfurl records in the database.

With this PR we responds to Slack's event immediately and then proceed to process the event. This pattern is also encouraged by the Slack API.

Unfortunately this pattern seems incompatible with our current integration testing setup

### DIFF
--- a/lib/unfurls/index.js
+++ b/lib/unfurls/index.js
@@ -6,10 +6,12 @@ async function linkShared(req, res) {
   const { event } = req.body;
 
   req.log.debug(req.body, 'Slack event received');
+  // Acknowledge receipt of event immediately
+  res.send();
 
   // if there are 3 or more, don't unfurl at all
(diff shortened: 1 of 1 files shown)


wilhelmklopp:
@bkeepers any thoughts on how we could adjust the unfurl tests so that it doesn't rely on the request/response cycle of an `await request(probot.server).post()...`

bkeepers:
Hmm, I'm not really sure how we could make that work. The tests need a callback or promise that they can wait for to know when the logic under test is done and assertions can be made.

PR #533, opened 2018-04-16
Add slack metadata to support message

wilhelmklopp:
As suggested in #527, this PR adds Slack metadata to the support message which will make it easier to troubleshoot problems

image

cc @izuzak


izuzak:
Thanks for adding this! Do you think it would be okay to also add the actual command the user ran and which triggered the error? 💭

wilhelmklopp:
> Thanks for adding this! Do you think it would be okay to also add the actual command the user ran and which triggered the error? 💭

Yes! Great idea :)

Issue #52, opened 2017-10-24
Unfurl: private links


bkeepers:
A user sees an ephemeral prompt to verify that the private link should be unfurled (always, this time, never)
A user sees an unfurl of a private link.

**Estimate:** 2 weeks

## Private unfurls are available via EARLY ACCESS
Read through this comment to find out how to get early access: https://github.com/integrations/slack/issues/52#issuecomment-387410861

## Road to private unfurls
Things that need to be done before general availability:
- [x] Persistent user settings #528
- [x] Better flow in case a user has not yet linked their GitHub account (building on top of #536)
- [x] Let users change settings using `/github settings`
- [ ] dm user if chat.ephemeral doesn't work because app is not in channel
     - [ ] ~Can we prompt users via additive permissions to add GitHub to the channel?~ Nope, not really :/
- [ ] Remove early access for private unfurls :tada:

olivierlacan:
I was a bit surprised when after subscribing a channel to a private repo, private URLs to code from that same repo weren't unfurled. That seemed like what this suggested: 

![image](https://user-images.githubusercontent.com/65950/36686685-815f01f6-1b27-11e8-80b6-3a0ebb75cb40.png)

It seems like until this feature is developed, the site should be updated to say "Public code snippets" to clarify. I wouldn't surprised if a lot of people made the same assumption I did and show up here to complain. 😄

brennaheaps:
@olivierlacan Hi! 👋

bkeepers:
Thanks @olivierlacan. The copy did say "public" at one point, but it must have gotten dropped in one of the iterations.

Private unfurls are coming very soon.

krumware:
+1 for confusing language, expected private repos to work. 
Regardless, excited about this coming feature, thanks folks!

bradical:
I had a question about how this will ultimately function. Will Private Repository unfurls depend on being subscribed to that repository in the channel or will it just use the perms of the authorizing user to determine access to the URL and could use any URL for any repo regardless of channel subscriptions?

wilhelmklopp:
> Will Private Repository unfurls depend on being subscribed to that repository in the channel or will it just use the perms of the authorizing user to determine access to the URL and could use any URL for any repo regardless of channel subscriptions?

It will most likely be based on the access of the user who posts the link. So any URL for any repo will work, but there may be an additional step where you need to confirm that you do want the link unfurled.

bradical:
Thank you! Looking forward to this.

[...]

PR #536, opened 2018-04-17
Private unfurls: Prompt users to connect their GitHub account

wilhelmklopp:
At the moment we just throw an error (and show nothing to the user) when a private unfurl fails due the user not having connected their GitHub account.

As part of extending the early access for private unfurls, we should prompt users to connect their GitHub account, otherwise they won't know why private unfurls aren't working.

With this PR we send a simple ephemeral message to the user for whom the private unfurl failed. Ideally this message itself should contain the "Connect GitHub account" link-button, but that should only be done once the command/PendingCommand refactor is complete.


lib/unfurls/index.js
```diff
@@ -53,10 +54,19 @@ async function linkShared(req, res) {
...
+          user: event.user,
+          attachments: [new PromptSignIn().getAttachment()],
+        });
+        return Promise.resolve();
+      }
       throw err;
     }
     if (hasEarlyAccess({ channelId: event.channel, teamId: req.body.team_id })) {
       if (newUnfurl && !newUnfurl.isDelivered) {
+        // todo: If app is not invited to channel, then send as DM to user
```
bkeepers: What if the app didn't attempt to unfurl if it's not invited to the channel?

Either way, would it make sense to add this as a checkbox on #52 instead of an inline TODO?

wilhelmklopp: > What if the app didn't attempt to unfurl if it's not invited to the channel?

We don't have a way of knowing whether the app is in the channel until we try to send an ephemeral message, from what I can tell :/ 
Unfurling works fine even if the app is not in the channel

> Either way, would it make sense to add this as a checkbox on #52 instead of an inline TODO?

Yes, good point 👍 I'll add it as a comment #528

PR #538, opened 2018-04-17
API to post messages to Slack

### LINKED ISSUE #478: API to post messages to Slack
No matter how many features we build into the GitHub + Slack integration, there are going to be use cases and workflows we don't support. Someone somewhere will want something related to their work on GitHub to show up in their Slack channel that may not be broadly applicable, and they don't want to implement their own Slack integration just to make it work.

It would be awesome if the could leverage the fact that GitHub already integrates with Slack and post messages into any channel subscribed to a specific repository.

```
curl -xPOST -H "Authorization: token $token" \
   https://slack.github.com/repos/:owner/:repo/:type
   -d "$message"
```

Where:

- `$token` is any valid GitHub token (personal access token, OAuth token, GitHub App token) that has write access to `:owner/:repe`
- `:type` is a user-defined type for the message (like `tickets`) that can be subscribed/unsubscribed to in the channel (`/github subscribe myorg/myrepo tickets`)
- `$message` is any valid payload for a [slack message](https://api.slack.com/docs/messages)

bkeepers:
```
$ curl -H "Authorization: token $ANY_GITHUB_TOKEN" \
    -XPOST http://localhost:3000/repos/integrations/slack \
    -d '{"text": "Hello GitHub+Slack API"}'
{"ok":true}
```

screen shot 2018-04-17 at 8 19 40 am

TODO:
- [x] Use token from `Authorization` header and verify that it has _write_ access to the repository
- [ ] Allow posting with a GitHub App token
- [x] ~~Add an optional `:type` parameter (like `tickets`) that can be subscribed/unsubscribed to in the channel (`/github subscribe myorg/myrepo tickets`)~~ I'm going to skip this for now since it'll require some changes to how we validate subscription settings. We can add it in the future after we verify that this is a feature we want to support

cc #478


lib/api.js
```diff
@@ -0,0 +1,56 @@
...
+    // Lookup subscriptions for repository
+    const subscriptions = await Subscription.lookup(repository.id);
+
+    // Deliver message to each channel
+    await Promise.all(subscriptions.map(async (subscription) => {
+      const slack = subscription.SlackWorkspace.client;
+
+      const params = {
+        channel: subscription.channelId,
+        ...req.body,
```
wilhelmklopp: should we restrict this a bit more? So that only attachments or text are passed?

lib/api.js
```diff
@@ -0,0 +1,56 @@
...
+  app.use(bodyParser.urlencoded({ extended: true }));
+
+  app.post('/:owner/:repo', async (req, res) => {
+    const { owner, repo } = req.params;
+    const github = new GitHub({ logger: req.log });
+
+    // Ensure credentials were provided
+    if (!req.headers.authorization) {
+      res.status(401).send({ message: 'Requires authentication' });
+      return;
```
wilhelmklopp: nit: `return` could be on the line above

lib/api.js
```diff
@@ -0,0 +1,56 @@
...
+      const params = {
+        channel: subscription.channelId,
+        ...req.body,
+      };
+
+      req.log(params, 'Delivering message to subscription');
+      await slack.chat.postMessage(params);
+    }));
+
+    res.send({ ok: true });
```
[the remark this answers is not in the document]
bkeepers: I was wondering that too. The only thing I was wondering is what we do if there are multiple subscriptions for a repository.

bkeepers:
I added a checklist to #478. This currently works for personal access tokens, and I'll open another PR to add support for any other app.

PR #543, opened 2018-04-18
Handle permanent Slack errors

### LINKED ISSUE #197: Disable Subscriptions when a channel is archived or deleted
Currently seeing [this error](https://sentry.io/github-integrations/slack/issues/421500015/) in production:

```
Response not OK:  is_archived
```

That is because I set up a channel to test some subscriptions and then archived it before removing the subscriptions.

bkeepers:
Some errors that are currently being raised when delivering notifications are permanent. This rescues them and deletes the subscription so we stop attempting to deliver in the future.

Fixes #197 
Fixes #188

### DIFF
--- a/lib/activity/router.js
+++ b/lib/activity/router.js
@@ -1,6 +1,7 @@
 const { Subscription, SlackUser, GitHubUser } = require('../models');
 const { ReEnableSubscription } = require('../messages/flow');
 const avoidReplicationLag = require('../github/avoid-replication-lag');
+const isPermanentError = require('../slack/is-permanent-error');
 const cache = require('../cache');
 
 // Temporary "middleware" hack to look up routing before delivering event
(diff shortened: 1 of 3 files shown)


lib/activity/router.js
```diff
@@ -58,7 +58,21 @@ module.exports = function route(callback) {
         // Delay GitHub API calls to avoid replication lag
         context.github.hook.before('request', avoidReplicationLag());
 
-        return callback(context, subscription, slack);
+        try {
+          await callback(context, subscription, slack);
+        } catch (err) {
+          const PERMANENT_ERRORS = [
```
wilhelmklopp: Should `is_archived` and `token_revoked` also be included in this list?
From: https://api.slack.com/methods/chat.postMessage

Issue #446, opened 2018-03-09
Disable subscriptions when user revokes access to the slack app


bkeepers:
We're seeing a lot of `401 Bad Credentials` errors in the permission check that we run when delivering subscriptions.

Copying from https://github.com/integrations/slack/pull/444#issuecomment-371680380:

> After testing this in production, I'm 99% certain that the 401s are coming from users who have revoked access from the GitHub App. I think the correct behavior is to treat a 401 as a 404 and disable the subscription.
>
> We could do a custom error message for 401 that says something about the user revoking access.

cc https://sentry.io/github-integrations/slack/issues/473125081/events/15502603286/

bkeepers:
There's a new `github_app_authorization` webhook that is [staff shipped](https://github.com/github/github/pull/86624) for our app:

```
{
  "action": "revoked",
  "sender": {
    "login": "octocat",
    "id": 1,
    "avatar_url": "https://github.com/images/error/octocat_happy.gif",
    "gravatar_id": "",
    "url": "https://api.github.com/users/octocat",
    "html_url": "https://github.com/octocat",
    "followers_url": "https://api.github.com/users/octocat/followers",
    "following_url": "https://api.github.com/users/octocat/following{/other_user}",
    "gists_url": "https://api.github.com/users/octocat/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/octocat/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/octocat/subscriptions",
    "organizations_url": "https://api.github.com/users/octocat/orgs",
    "repos_url": "https://api.github.com/users/octocat/repos",
    "events_url": "https://api.github.com/users/octocat/events{/privacy}",
    "received_events_url": "https://api.github.com/users/octocat/received_events",
    "type": "User",
    "site_admin": false
  }
}
```

PR #545, opened 2018-04-18
/github signout

wilhelmklopp:
Add `/github signout` as an undocumented command to help with discovering edge cases and as an option in case users accidentally? link the wrong GitHub account

Todo:
- [x] Get pending command to not loop infinitely
- [x] add tests
- [x] Figure out other implications of sign out. Do we delete subscriptions automatically?
- [x] delete subscriptions
- [ ]~post interactive message with 'confirm' option. Also have an option that includes: 'Post message in each channel prompting to re-enable all disabled subscriptions'~ not doing this right now because of the effort needed to make it happen
- [x] make signing out a method on the model so that it can be re-used when initiated by `github_app_authorization.revoked` in https://github.com/integrations/slack/issues/446


lib/commands/signout.js
```diff
@@ -0,0 +1,23 @@
+const { SignOut } = require('../messages/flow');
+
+const SignedParams = require('../signed-params');
+
+module.exports = async (req, res) => {
+  await res.locals.slackUser.update({
```
bkeepers: Nit: since there are a few `res.locals` here, wdyt about pulling them out as variables first so it's obvious what dependencies this has?

lib/commands/signout.js
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  const { command } = res.locals;
+
+  const state = new SignedParams({
+    trigger_id: command.trigger_id,
+  });
+
+  const protocol = req.headers['x-forwarded-proto'] || req.protocol;
+  const host = req.headers['x-forwarded-host'] || req.get('host');
```
bkeepers: What do you think about pulling this out into a module? I think we do this in 2-3 places now.

lib/messages/flow/signout.js
```diff
@@ -0,0 +1,28 @@
...
+
+  toJSON() {
+    return {
+      response_type: 'ephemeral',
+      attachments: [{
+        ...this.getBaseMessage(),
+        fallback: 'Your GitHub account has been disconnected.',
+        text: `:white_check_mark: <@${this.userSlackId}> is disconnected from your GitHub account\n` +
+          'Features like subscriptions and unfurls will stop working. ' +
+          'To re-connect your GitHub account click the button below.',
```
bkeepers: I know we're already using  "connect" elsewhere, but what about switching to the language of "signed in" and "signed out". I also don't think we've been using "unfurls" in any user-facing language.

> :white_check_mark: <@U98765> is now signed out. Use `/github signin` to sign back in to your GitHub account at any time.

🤷‍♂️ 

@cmwinters, what do you think?

cmwinters: @bkeepers seems fine to me.

wilhelmklopp: 👍 Also prefer to avoid "connect". I think it is valuable to tell the user that features will stop working though. That might not be obvious to them.

lib/commands/signout.js
```diff
@@ -0,0 +1,44 @@
...
+      const subscriptions = subscriptionsByChannel[channel];
+      const repositories = await getRepositories(subscriptions, robot);
+      if (repositories) {
+        const attachment = new ReEnableMultipleSubscriptions(repositories, slackUser.slackId, 'disconnected their GitHub account').getAttachment();
+        return slackWorkspace.client.chat.postMessage({
+          channel,
+          attachments: [attachment],
+        });
+      }
+    }));
```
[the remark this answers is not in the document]
wilhelmklopp: Extra context: I plan to refactor this code, so that it will be used both by `/github signout` AND the code for the "token revoked" event. That event will probably lead to this code path being called much more frequently: https://github.com/integrations/slack/issues/446

In a way this PR was just a set up for this follow up one

PR #549, opened 2018-04-21
subscribe just to tag activity

philschatz:
This adds the ability to subscribe _only_ to tags being created or deleted and not include when a branch is created or deleted.

The use-case we have for this feature is that our QA team would like to know when new versions of our code are tagged by developers so they can start their tests.

I am not sure if this is the best way to implement the feature since the actual webhook event is a [`create`](https://developer.github.com/v3/activity/events/types/#createevent), not a `branch` event but if something like this is useful I can try to change it to be `tags` instead of `branches:tag`.


README.md
```diff
@@ -106,6 +106,7 @@ These are disabled by default, and can be enabled with the `/github subscribe ow
 - `reviews` - Pull request reviews
 - `comments` - New comments on issues and pull requests
 - `branches` - Created or deleted branches
+- `branches:tag` - Created tags
```
bkeepers: From an implementation perspective I think it makes sense to include this on `branches`, but from a user perspective, would make sense to include this as a top-level setting? `branches:tag` feels weird, whereas `/github subscribe owner/repo tags` makes a little more sense to me.

PR #554, opened 2018-04-24
Open Issue from slack

wilhelmklopp:
Implement opening issues from Slack, using dialogs 🎉 

TODO:
- [x] tests
- [x] move to lib/messages

Depends on #528 
Depends on new GitHub App permissions


lib/create/index.js
```diff
@@ -0,0 +1,126 @@
...
+  });
+
+  return res.send();
+}
+
+
+module.exports = {
+  openDialog,
+  dialogSubmit,
+};
```
wilhelmklopp: This module is called `create` because I couldn't come up with a better name. 
The idea is that you can use the `/github new x` syntax to create different things. Right now: issues, in the future: team discussions, and others

lib/create/index.js
```diff
@@ -0,0 +1,73 @@
...
+    gitHubUser,
+    slackWorkspace,
+  } = res.locals;
+
+  const subscriptions = await Subscription.findAll({
+    where: { channelId: req.body.channel_id, slackWorkspaceId: slackWorkspace.id },
+  });
+  if (subscriptions.length === 0) {
+    return command.respond({ text: 'This channel is not subscribed to any repositories' });
+  }
```
bkeepers: It'd be great if a user could also pass in an explicit repo here. `github open owner/repo`

lib/create/index.js
```diff
@@ -0,0 +1,73 @@
...
+      return repository.data;
+    } catch (err) {
+      req.log.error({ err, repoId: subscription.githubId }, 'Could not find repository for subscription');
+      if (err.code !== 404) {
+        throw err;
+      }
+    }
+  }));
+  // remove undefined
+  repositories = repositories.filter(repo => repo);
```
bkeepers: I think we're fetching a list of repositories elsewhere. Can we move this logic into a reusable module somewhere that takes a client and a list of repositories IDs and returns the repos?

lib/create/index.js
```diff
@@ -0,0 +1,73 @@
...
+  return command.respond();
+}
+
+async function dialogSubmit(req, res) {
+  const { slackUser, slackWorkspace } = res.locals;
+
+  const { title, body } = req.body.submission;
+  let { repository } = req.body.submission;
+
+  if (!repository) {
```
bkeepers: When will this not be set?

lib/create/index.js
```diff
@@ -0,0 +1,73 @@
...
+    const subscription = await Subscription.findOne({
+      where: { channelId: req.body.channel.id, slackWorkspaceId: slackWorkspace.id },
+    });
+    repository = subscription.githubId;
+  }
+  await slackUser.reload({ include: [GitHubUser] });
```
bkeepers: I think this should already be loaded: https://github.com/integrations/slack/blob/master/lib/middleware/authenticate.js#L13

lib/create/index.js
```diff
@@ -0,0 +1,73 @@
...
+  const { data } = await slackUser.GitHubUser.client.repos.getById({ id: repository });
+  const owner = data.owner.login;
+  const repo = data.name;
```
bkeepers: ugh, makes me sad that this is necessary.

lib/messages/create/create-issue-dialog.js
```diff
@@ -0,0 +1,38 @@
...
+      label: 'Repository',
+      type: 'select',
+      name: 'repository',
+      placeholder: 'Select the repository on which to open the issue',
+      options: repositories.map(repository => ({
+        label: repository.full_name,
+        value: repository.id,
+      })),
+    });
+  }
```
bkeepers: For a future iteration it'd be cool to power this with an external data source so it can be dynamic and use any repo that the user has access to: https://api.slack.com/docs/message-menus

lib/create/index.js
wilhelmklopp: Do you think we could add that in a follow up PR?

lib/create/index.js
wilhelmklopp: When there is just one repo subscribed in the channel, then we default to that one

lib/messages/create/create-issue-dialog.js
wilhelmklopp: Yeah, that'd be neat! I wonder what endpoint/graphQL query we should use to fetch the most relevant repos

lib/create/index.js
wilhelmklopp: The `authenticate` middleware isn't actually at play here because this line is an "action route", so it only passes through this middleware, which doesn't load the GitHub user: https://github.com/integrations/slack/blob/master/lib/middleware/action-attach-metadata.js

lib/create/index.js
wilhelmklopp: Having said that I wonder if we could have a message menu with just one repository. That might be more consistent both for users and in the implementation 👍

PR #561, opened 2018-04-28
Add /github settings

wilhelmklopp:
Implements `/github settings` to remove repos from auto-unfurls and unmute muted prompts

image

Todo:
- [x] Integration tests
- [x] -> /lib/messages
- [x] Add entry in `/github help`


ppamorim:
Is this working? When I do `/github settings` it returns `You do not have any settings configured.`. How to add a configuration?

wilhelmklopp:
@ppamorim what setting would you like to configure?

creichert:
Mine says the same thing `You do not have any settings configured.`. Really, just a couple examples would be nice. I don't know what settings I would like to configure yet. For new issues created, I would like the title of the issue in the desktop notification (is that possible like in the GitHub legacy slack app)?

wilhelmklopp:
`/github settings` is currently only for individual user settings relating to unfurls. 

Examples for repo level subscription configurations are here: https://github.com/integrations/slack#configuration

> For new issues created, I would like the title of the issue in the desktop notification (is that possible like in the GitHub legacy slack app)?

I don't think it'll currently show the title of the issue. Only something to the effect of "[integrations/slack] Issue opened by creichert" -- but feel free to open an issue on this repo for this improvement

creichert:
@wilhelmklopp Perfect, thanks!

matkoch:
The thing is that `/github help` reveals `settings` as a command, but after invoking `/github settings` this doesn't give the user any clue of what it can do. More documentation would be really appreciated.

say8425:
@matkoch I agree. Also this `settings` command is not for most of users. So I think change the command name.

nevir:
How do we use this command to adjust unfurl settings?

PR #582, opened 2018-05-16
 DM user if app cannot post ephemeral private unfurl prompt

wilhelmklopp:
If the workspace token app has not yet been 'added to the conversation' to a channel, then folks won't know about the private unfurls feature.

This PR introduces logic where we DM them instead with the same prompt, and also ask them to invite `@github` to the channel where they just shared the link.

Todo:
- [x] integrations tests
- [x] also include SignIn prompt
- [x] make the follow up prompts work in the way that a user would expect (if possible)


wilhelmklopp:
Closing this, because it would go against recent ideas of a channel-based "banlist" where you do not want private unfurls to occur.

PR #585, opened 2018-05-17
Open issue from Slack by specifying repo

wilhelmklopp:
Replaces #554 

Channels with lots of subscriptions are not able to use the implementation in #554 so changing it almost entirely to take a repository as an input argument

Instead of creating an issue with: `/github new issue`
You know have to specify: `/github open owner/repo`


steveoh:
Would you doc these new features in the readme and in the slack `/github` command? They both still only reference the subscribe actions. 

PS the slack.github.com page shows `/example` when I think you meant to replace that with an actual example.

image

bkeepers:
Hey @steveoh, thanks. The docs were updated in #586 (after your comment here).: https://github.com/integrations/slack#take-action

steveoh:
thanks @bkeepers how about the slash command in slack? It seems like it needs to default to `/github help` that could print the usage. Now it only shows subscribe... 

Just a though.

PR #587, opened 2018-05-17
Verify permissions for actions

bkeepers:
Some features we're rolling out now will require updated permissions. Account owners will receive an email prompting them to accept the new permissions, but other users may attempt to use the new features before admins accept the permissions.

This will show a message if permissions are not updated when a user tries to use the new feature (like `/github close owner/repo#123`:

slack_-_github_slack_dev

TODO:
- [x] add middleware to verify permissions
- [x] Use on `close`/`reopen` commands
- [x] Use on `open` dialog
- [x] Use on `/deploy` commands
- [ ] tests


Cervator:
Got the "Slack is requesting updated permissions" email. Got confused about what triggered it. Started digging around here, became enlightened. Thought this might be the most relevant place to post a quick note. Keep up the good work 👍 

Would be really nice if you (in general) could add in a comment when triggering such an update from an existing GitHub app, or a link to an associated PR / blog post / etc :-)

bkeepers:
> Would be really nice if you (in general) could add in a comment when triggering such an update from an existing GitHub app, or a link to an associated PR / blog post / etc :-)

I agree @Cervator. We've given this feedback to the GitHub apps team.

lib/messages/flow/github-app-permission-update.js
```diff
@@ -0,0 +1,30 @@
...
+  get guidance() {
+    if (this.owner.id === parseInt(this.gitHubUser.id, 10)) {
+      return '';
+    }
+    if (this.owner.type === 'User') {
+      return '\n_Note: You will need to ask the owner of the repository to install it for you. ' +
+      `Give them <${this.installLink}|this link.>_`;
+    }
+    return '\n_Note: You need to be an organization owner to install the app (or ask one to install it for you)._';
+  }
```
wilhelmklopp: looks like guidance is not used?

wilhelmklopp:
We'll probably want to use the assertPermission method on the `Installation` model now that #606 has landed: https://github.com/integrations/slack/blob/4d2bd707aa26ea4af188313ca710891c41730944/lib/models/installation.js#L62-L98

IAmHughes:
Reopening this. @bkeepers thank you for all your work on this! I'm going to piggy back off this code/branch (so you get credit too!!) and hopefully work to merge this in 👍

IAmHughes:
I've been working on this pushing up code trying to get things working as expected. I need to add at least one more additional test (if not a couple) but things are working!

I wrapped each slash command (`commands/index.js`) in a check for permissions, but we ONLY check the permissions if we receive a 403 or 404 from the GitHub API. That way, we don't check for proper permissions EVERY time like the original route was taking us.

IAmHughes:
Got the permission check working for `/github close` and `/github reopen` of issues. Need to work on the `/github open` command as it's handled differently (with popup dialog).

image

_Note:_ IAmSlackBotHughes is my dev environment's bot 😄 🤖

IAmHughes:
Update!

This now works on the `/github open`, `/github close`, and `github reopen` commands as well as the `/github deploy` command!

I also updated the way we represent the error message so that we can provide a link that can be sent to the owner/admin if needed:

image

IAmHughes:
Status Update: This is almost ready, working on tests for some new functionality and correcting snapshots now that we are adding some addition information in Slack dialogs (specifically `state`).

lib/create/deployment.js
```diff
@@ -17,42 +32,92 @@ async function open(req, res) {
...
   const { owner, repo } = resource;
+
   const [{ data: repository }, { data: branches }, { data: tags }] = await Promise.all([
     gitHubUser.client.repos.get({ owner, repo }),
     gitHubUser.client.repos.listBranches({ owner, repo }),
-    gitHubUser.client.gitdata.getTags({ owner, repo }),
+    getTags(gitHubUser, owner, repo),
   ]);
-
+  console.log('tags', tags);
```
gimenete: Whoops, we should remove this.

lib/dialogs/add-state.js
```diff
@@ -0,0 +1,4 @@
+module.exports = (dialog, channel, resource) => {
```
gimenete: Can we add a comment here explaining what this function is for? Something like:

> This function adds a `state` property for dialogs, in order to carry additional, hidden, information to the webhook received when the dialog is submitted. Typically this information will be used in middleware functions and injected into attributes of `res.locals`.

lib/create/deployment.js
IAmHughes: Haha, oops, fixed that!

lib/dialogs/add-state.js
IAmHughes: Good call!

IAmHughes:
This one is ready to go! Thank you @gimenete and @wilhelmklopp for your help along the way.

If I could have a final 👍 on this, I'll be deploying to staging today and then production if all goes well.

/cc @eekdageek @imjohnbo for awareness.

PR #589, opened 2018-05-17
fail on missing STORAGE_SECRET

andymckay:
Add in notes to the docs about the storage secret. Also ensure that if its not specified instead of getting the error below, we get a quick fail:

```
7:18:00.182Z ERROR http: Invalid key length (id=4bc6e05b-f6cf-4453-b513-f8f6dc270ade)
  Error: Invalid key length
      at new Cipheriv (crypto.js:219:16)
```

### DIFF
--- a/lib/models/github-user.js
+++ b/lib/models/github-user.js
@@ -5,7 +5,7 @@ const Sequelize = require('sequelize');
 const EncryptedField = require('sequelize-encrypted');
 
 if (!process.env.STORAGE_SECRET) {
-  throw 'STORAGE_SECRET is not defined.'
+  throw new Error('STORAGE_SECRET is not defined.');
 }
 
 const encrypted = EncryptedField(Sequelize, process.env.STORAGE_SECRET);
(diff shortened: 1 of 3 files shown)


lib/models/github-user.js
```diff
@@ -4,6 +4,10 @@ const logger = require('../logger');
 const Sequelize = require('sequelize');
 const EncryptedField = require('sequelize-encrypted');
 
+if (!process.env.STORAGE_SECRET) {
+  throw 'STORAGE_SECRET is not defined.'
```
bkeepers: `throw new Error('STORAGE_SECRET is not defined.');` should fix both of these lint failures:

```
  8:3   error  Expected an object to be thrown  no-throw-literal
  8:41  error  Missing semicolon                semi
```

lib/models/slack-workspace.js
```diff
@@ -3,6 +3,10 @@ const EncryptedField = require('sequelize-encrypted');
 
 const slack = require('../slack/client');
 
+if (!process.env.STORAGE_SECRET) {
+  throw 'STORAGE_SECRET is not defined.'
```
bkeepers: Same here

Issue #557, opened 2018-04-25
Allow subscribing to multiple repos?


hugovk:
It would be great to be able to do something like:

```
/github subscribe owner/repo1 owner/repo2 owner/repo3 [feature]
```

(Also unsubscribe.)

This has been mentioned in the comments of https://github.com/integrations/slack/issues/391, but here's its own an issue as a feature request.

We have a Slack where most of our repos are in #general, but some channels are for single repos, and some channels are for a set of repos.

Mass subscribing would allow people to add just those repos they want to a specific channel.

It also allows us to more easily unsub a batch of repos from #general and re-sub in a new channel.

This could also allow changing subscriptions of features (eg. `issues`, `pulls`).

Thank you!

ahyattdev:
This! At the Darling Project, we have around 100 repos. `/github subscribe darlinghq/*` to subscribe to all of them would be nice!

hugovk:
@stale Still relevant, keep open. I've not tested it, but perhaps https://github.com/integrations/slack/issues/602 will fix it?

nicwortel:
Yes, this is still relevant.

ahyattdev:
Would very much like to see this. Our GitHub organization has about 100 repos and subscribing to them all is a pain.

Vmadmax:
Any progress?

vircea:
Any progress?

SimonTouitou:
Why is this still open?
Have a look at: https://github.com/integrations/slack/issues/391
Resolved long ago.

Quote:

> I'm happy to announce that we have just deployed this feature through #709
> 
> Use:
> 
> > /github subscribe Org-name
> 
> Be aware that if you are already subscribed to repositories in that org you will get duplicated notifications depending on what you are subscribed to in those repositories and the organization. Remember that you can subscribe just to certain things (pulls, commits, issues, etc). Example:
> 
> Subscribe to everything
> 
> > /github subscribe owner/repo
> 
> Then unsubscribe for everything except for issues
> 
> > /github unsubscribe owner/repo commits,pulls,deployments,statuses,public,releases
> 
> You can list your current subscriptions with
> 
> > /github subscribe list
> 
> Enjoy it! 🎉

psafont:
> Why is this still open?
An organization can have multiple teams, each one taking care of many repositories inside the same organization.

[...]

Issue #391, opened 2018-02-27
Allow subscribing to an entire organization's collection of repos?


rbonestell:
I have recently upgraded from the old GitHub slack plugin to the new one, and my organization has almost 80 repositories. I don't want to have to copy/paste the link and subscription type for each of my repositories into each channel I want the bot in. Additionally, consider auto-subscribing to new repos in an organization given a specific configuration?

erydo:
@bkeepers — As an additional data point, our org also has >80 repositories. We previously had a single Slack/GitHub webhook set up at the org level. It was very easy.

Since the new plugin doesn't even let you subscribe to multiple repositories in the same message, e.g.:
```
/github subscribe foo/bar
/github subscribe foo/baz
```

The message for each of those had sent individually. (First, I had to figure out how to list all our repositories…). It was an extremely annoying migration, and now I'm not confident we'll see messages on newly created repositories.

DuBistKomisch:
This is desperately needed. Just migrated almost 30 repos with copy pasta and then found out I have to go back through and enable all the PR notifications for each one!

vbezruchkin:
Hey guys,

Thanks for the integration.

Also 80+ public repositories for our organization.. And it's not easy to subscribe to each one separately. 
We would like to use this app instead of hook for an organization.
Any ETA?

Thanks

danrademacher:
A halfway point that would help would be to allow an array of repos to be submitted with one command. At least then one could construct a single command to batch subscribe to repos.

bkeepers:
> Any ETA?

Sorry, not yet. I know the pain because I have a couple organizations that have tons of repos. We'll be switching into new feature mode soon and will take this into consideration when planning.

chriszs:
To do this with the old integration, I set a webhook for my organization under organization Settings -> Webhooks. This still works for now, but I don't see any benefits of the new integration, so I assume I'm still on the old one.

weisjohn:
Is there a way to migrate from the new integration to the old?

esaruoho:
+1 for proper migration from old-legacy-github to nu-github without losing comments/pullrequestreviews etc.

chriszs:
[...]

Issue #56, opened 2017-10-24
Configuration: /subscribe owner


bkeepers:
`/github subscribe owner` to subscribe to notifications for an account

**Estimate:** 0.5 weeks

anowlcalledjosh:
This would be awesome to have, and a clear improvement over the "legacy" GitHub integration :)

anowlcalledjosh:
As before, this would still be a lovely feature to have.

wilhelmklopp:
This has shipped!

PR #602, opened 2018-05-24
add possibility to subscribe to an entire organization

GertSallaerts:
This is related to multiple issues: https://github.com/integrations/slack/issues/56, https://github.com/integrations/slack/issues/391 and slightly less relevant https://github.com/integrations/slack/issues/557

It proved to be a more challenging issue than I suspected at first, so I went for a basic implementation to get started. I'd be interested to know what you guys think (besides the fact that extra tests should probably be added 🙂.)

__"New" commands:__

- Subscribing `/github subscribe owner`
- Unsubscribing: `/github unsubscribe owner`
- Settings: `/github subscribe owner comments`

### DIFF
--- a/lib/commands/list-subscriptions.js
+++ b/lib/commands/list-subscriptions.js
@@ -21,28 +21,40 @@ module.exports = async (req, res, next) => {
     where: { channelId: command.channel_id },
   });
 
-  let repositories = await Promise.all(subscriptions.map(async (subscription) => {
+  let resources = await Promise.all(subscriptions.map(async (subscription) => {
     const github = await robot.auth(subscription.Installation.githubId);
+    let githubResource;
+
     try {
-      const repository = await github.repos.getById({ id: subscription.githubId });
-      return repository.data;
+      githubResource = await github.repos.getById({ id: subscription.githubId });
+      return githubResource.data;
     } catch (err) {
-      req.log.error({ err, repoId: subscription.githubId }, 'Could not find repository for subscription');
       if (err.code !== 404) {
         throw err;
       }
     }
+
+    try {
+      githubResource = await github.users.getById({ id: subscription.githubId });
+      return githubResource.data;
+    } catch (err) {
+      if (err.code !== 404) {
+        throw err;
+      }
+    }
+
+    req.log.error({ githubId: subscription.githubId }, 'Could not find resource for subscription');
   }));
 
   // remove undefined
-  repositories = repositories.filter(repo => repo);
+  resources = resources.filter(repo => repo);
 
   if (command.text === 'list') {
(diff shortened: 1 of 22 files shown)


justinperkins:
Would this include subscribing to all branches (existing and new) on all repos?

austingray:
What's the next step here? This is in demand! :)

ibarrajo:
This is very promising! Shame the PR is getting stale =/

Do you need any help with this? At this point I'm seriously considering writing an internal slack app just for the purpose of subscribing stakeholders to notifications related to them.

GertSallaerts:
Still committed to get this merged, but would need some feedback as to what is needed to get this up to snuff. I could get test coverage up to 100% but the lack of response seems to suggest it isn't that simple?

@ibarrajo Personally I'm not interested in maintaining a fork of this, but if you decide to go for an internal app, feel free to contact me if you need help with something.

ibarrajo:
@GertSallaerts Thank you! Let's hope your PR gets picked up soon. There are many people interested in seeing this functionality.

It's a non-trivial PR so I guess that it would take some engineering effort from Github to actually code review and approve.

lib/commands/list-subscriptions.js
```diff
@@ -21,28 +21,40 @@ module.exports = async (req, res, next) => {
...
+
+    try {
+      githubResource = await github.users.getById({ id: subscription.githubId });
+      return githubResource.data;
+    } catch (err) {
+      if (err.code !== 404) {
+        throw err;
+      }
+    }
+
```
wilhelmklopp: Rather than trying for both repository id and user id in this file, could we add another column in the `Subscriptions` table that indicates whether the subscription is to an "acccount" or to a "repo"?

lib/commands/list-subscriptions.js
```diff
@@ -21,28 +21,40 @@ module.exports = async (req, res, next) => {
...
-    return command.respond(new SubscriptionList(repositories, command.channel_id));
+    return command.respond(new SubscriptionList(resources, command.channel_id));
   }
 
   const response = (new Help(command.command, command.subcommand)).toJSON();
-  const list = new SubscriptionList(repositories, command.channel_id);
+  const list = new SubscriptionList(resources, command.channel_id);
   response.attachments.push(list.toJSON().attachments[0]);
   return command.respond(response);
 };
```
wilhelmklopp: I think I'd also prefer if for the `/github subscribe list` command we could list the number of repositories and the number of "accounts" as separate counts and lists as opposed to adding them together.

For what it's worth, I'm also not completely in love with the "resources" name here. Is there another name that we could use here?

GertSallaerts:
Completed in https://github.com/integrations/slack/pull/709

PR #605, opened 2018-05-25
Upgrade to Probot 7

bkeepers:
Probot 7 is a rewrite in TypeScript (https://github.com/probot/probot/issues/512). Since we're using some internal APIs, we have to change the way we're loading a few things.

This should be ready to go, but I'm going to let it sit until Probot 7.0 is officially released.

cc @tcbyrd


bkeepers:
> I suppose we don't need to use `context.name` as opposed to `context.event` wherever we access it explicitly?

No, we kept `context.event` around in Probot since `context.name` is kinda weird. The change from `event` to `name` was because `@octokit/webhooks.js` changed it. I suppose we should consider deprecating it in Probot too for consistency though. 🤔

PR #606, opened 2018-05-25
WIP [Message actions] Attach message to issue/pr thread

wilhelmklopp:


lib/models/installation.js
```diff
@@ -57,6 +58,43 @@ module.exports = (sequelize, DataTypes) => {
     async uninstall(payload) {
       return Installation.destroy({ where: { ownerId: payload.account.id } });
     },
+
+    async assertPermissions(github, { owner, repo, permissions }) {
```
wilhelmklopp: @bkeepers adding this function here already which is in may ways similar to #587  but different in the sense that permissions can be asserted outside of the middleware. Curious for your thoughts

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+      });
+    } else {
+      issuesByRepo[issue.repository.nameWithOwner] = [
+        {
+          label: formatLabel(issue),
+          value: issue.id,
+        },
+      ];
+    }
+  });
```
bkeepers: I would be great to extract all of this manipulation into a module/model of some sort.

lib/messages/create/comment-created.js
```diff
@@ -0,0 +1,17 @@
...
+module.exports = class CommentCreated extends Message {
+  constructor(commentUrl) {
+    super({});
+    this.commentUrl = commentUrl;
+  }
+
+  getAttachment() {
+    return {
+      ...super.getBaseMessage(),
+      text: `:white_check_mark: Comment successfully created: ${this.commentUrl}`,
```
bkeepers: Would it make sense to do this more in the form of the other activity messages where it's `${author} commented` and then the title of the issue/PR?

lib/middleware/require-github-user.js
```diff
@@ -0,0 +1,32 @@
...
+const { GitHubUser } = require('../models');
+const SignedParams = require('../signed-params');
+const getProtocolAndHost = require('../get-protocol-and-host');
+const { SignIn } = require('../messages/flow');
+
+/**
+ * Require the user to have connected their GitHub account
+ */
+module.exports = async function requireGitHubUser(req, res, next) {
+  const slackUser = await res.locals.slackUser.reload({ include: [GitHubUser] });
```
bkeepers: Is the `reload necessary`?  Can we just eager load `GitHubUser` wherever we're originally loading the slack user?

lib/middleware/route-option.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Routes the request based on action `type`, `callback_id`, and `name`
+ */
+module.exports = function routeOption(req, res, next) {
+  req.log.debug({ ...req.body, token: undefined }, 'Option load received from Slack');
+
+  req.url += `:${req.body.type}:${req.body.callback_id}:${req.body.name}`;
```
bkeepers: I don't think we need to do this now, but I'm starting to wonder if we should extract some kind of "parameter router" middleware instead of mutating the `req.url` everywhere.

Something like this:

```js
router.post('/actions', withParams({type: 'message_action', callback_id: 'attach-to-issue'}, (req, res) => {
  // route handler here
})
```

It's a little more verbose, but would be more generic/reusable.

lib/models/installation.js
bkeepers: I like having it in the model, and the command middleware can just reuse this.

lib/slack/index.js
```diff
@@ -60,7 +60,33 @@ module.exports = (robot) => {
...
 
+  router.post(
+    '/actions:message_action:attach-to-issue',
+    middleware.requireGitHubUser,
+    create.attachToIssueFromMessage.openDialog,
+  );
+  router.post(
+    '/actions:dialog_submission:add-comment',
+    middleware.requireGitHubUser,
+    create.attachToIssueFromMessage.createCommentDialogSubmission,
```
bkeepers: I'm not sure how many other dialogs we'll use, but it might be nice to standardize the naming of the methods to something django-ish/rails-y/resourceful. That way we can expect every dialog module has the same methods. Something like:

- `create.attachToIssueFromMessage.open`
- `create.attachToIssueFromMessage.submit`

README.md
```diff
@@ -96,6 +96,10 @@ Slack conversations often lead to decisions and actionable takeaways. Now it’s
 - Reopen an issue or pull request with `/github open [pull request link]`
 - Open a new issue with `/github open [owner/repo]`
 
+You can also take action on GitHub directly from a Slack message by clicking on the '`...` More Actions' menu available on every Slack message. From there you can:
+
+- Attach a message as a comment to an existing issue or pull request
```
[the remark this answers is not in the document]
wilhelmklopp: Right now the API kind of decides this for us. All we control is that the issues/prs relate to a user and are sorted by `UPDATED_AT` descending. Is that worth mentioning?

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+      issues: 'write',
+      pull_requests: 'write',
+    };
+    const { owner, repo } = resource;
+    if (!(await Installation.assertPermissions(await robot.auth(), { owner, repo, permissions }))) {
+      const appUrl = (await robot.info()).html_url;
+      return res.send({
+        errors: [{
+          name: 'url',
+          error: `Insufficient permissions. Please make sure the Slack app is installed on this repository: ${appUrl}`,
```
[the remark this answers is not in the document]
wilhelmklopp: > So this error will show up if I paste in a URL to a repository that doesn't have the app installed or hasn't granted write access?

Yep, that's correct. 

> Is the app URL clickable?

Unfortunately not :/
Looks like that's not supported using the native error handling for dialogs

> That feels a little unfriendly.

Agreed. Will update this

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+        }`;
+      data = (await slackUser.GitHubUser.client.query(
+        addComment,
+        { id: issueId, body: comment },
+      ));
+    }
+  } catch (err) {
+    if (!(err.name && err.name === 'GraphQLError')) {
+      throw err;
+    }
```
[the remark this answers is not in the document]
wilhelmklopp: We're catching the GraphQL error because it might be a permission issue. But we're already checking for permission using REST earlier (in case a user pasted a URL). So if there's an error as result of the REST API call, then it's a proper, 500-worthy error.

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+    req.log.debug({ err }, 'Could not create comment');
+
+    const appUrl = (await robot.info()).html_url;
+    return res.send({
+      errors: [{
+        name: 'issueId',
+        error: `Please make sure the Slack app has been granted issue and pull request write access on this repository: ${appUrl}`,
```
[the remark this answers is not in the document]
wilhelmklopp: Unfortunately we can only do the permissions check using the REST API (in the case where a user submits this dialog using the URL option. In case they select an issue (we have to use GraphQL) and there's no good way to figure out permissions of an installation using GraphQL right now (with less than one request).

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+      }],
+    });
+  }
+
+  let issueUrl;
+  if (data.html_url) {
+    issueUrl = data.html_url;
+  } else {
+    issueUrl = data.addComment.commentEdge.node.url;
+  }
```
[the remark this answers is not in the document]
wilhelmklopp: Yeah, I can give that a shot. Not sure how re-usable it will be because this dialog's way of creating a comment seems pretty unique (using either a URL and REST or a node_id and GraphQL), but at least we'll be able to reduce the size of this file that way.

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+          title
+          number
+          id
+          repository {
+            nameWithOwner
+          }
+        }
+      }
+    }
+  }`;
```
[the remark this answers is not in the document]
wilhelmklopp: Don't have an opinion on this, but yeah, we can do that 👍

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+
+  const { data } = await slackUser.GitHubUser.client.request({
```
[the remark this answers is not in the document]
wilhelmklopp: Yep, I'll add a comment. For posterity, it's because of the reasons described here: https://github.com/probot/probot/issues/484#issuecomment-393003393

lib/create/attach-to-issue-from-message.js
```diff
@@ -0,0 +1,219 @@
...
+    method: 'POST',
+    url: '/graphql',
+    query,
+  });
+
+  const { issues, pullRequests } = data.data.viewer;
+  let rawIssues = issues.nodes.concat(pullRequests.nodes).filter(issue => issue);
+
+  // If user supplied search term, then we filter based on the search term
+  if (value) {
```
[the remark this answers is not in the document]
wilhelmklopp: I understand that the Search API has not yet been enabled (or has been disabled) for User-to-Server requests, hence the roundabout approach of this implementation :/

lib/messages/create/comment-created.js
wilhelmklopp: Yeah, we could do that. Would we do this as an `in_channel` or `ephemeral` message? 

I suppose `in_channel` could be confusing to other users in the channel who did not invoke the message action. 

We should also only post this message using logic in this module if the repo and channel are not subscribed to comments anyway

lib/middleware/route-option.js
wilhelmklopp: Yeah, I like that a lot! Makes it much easier to understand what's going on especially for folks new to the codebase

lib/messages/create/comment-created.js
wilhelmklopp: Just looked into this, and unfortunately we don't get the issue/pr title or repository information from the create comment endpoint: https://developer.github.com/v3/issues/comments/#create-a-comment

We'd need that information to re-use the existing rendering logic. We don't really have an opportunity to make additional API calls before the dialog times out. A queue system would help here.

README.md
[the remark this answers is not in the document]
wilhelmklopp: Added another line for this: https://github.com/integrations/slack/pull/606/files#diff-04c6e90faac2675aa89e2176d2eec7d8R102

lib/dialogs/add-comment.js
```diff
@@ -0,0 +1,38 @@
+
+module.exports = (message, channel, team) => {
+  const linkToMessage =
+    `https://${team.domain}.slack.com/archives/${channel.id}/p${message.ts.replace('.', '')}`;
+
+  const comment = `> ${message.text}\n<sub>[View message in Slack](${linkToMessage})</sub>`;
```
bkeepers: > View message in Slack

What do you think about just `View in Slack`?

I also wouldn't be opposed to getting rid of the `<sub>` to make the dialog that appears in Slack a little cleaner, but don't feel strongly about it.

lib/dialogs/add-comment.js
```diff
@@ -0,0 +1,38 @@
...
+
+  return {
+    callback_id: 'add-comment',
+    title: 'Comment on Issue or PR',
```
bkeepers: I know we had a long discussion about not using `PR`, but don't remember the conclusion. Could this be `Comment on GitHub`?

lib/dialogs/add-comment.js
```diff
@@ -0,0 +1,38 @@
...
+        min_query_length: 0,
+        optional: true,
+      },
+      {
+        type: 'text',
+        subtype: 'url',
+        label: 'Issue/pull request URL',
+        placeholder: 'https://github.com/atom/atom/issues/1',
+        name: 'manualUrl',
+        hint: 'If you cannot find the issue or pull request in the menu above, you can paste its URL here instead.',
```
bkeepers: I usually try to avoid `here` when describing what to do (e.g. `click here`). Also, should we use `link` instead of `URL` to be a little friendlier?

What about this?

> A link to a different Issue or Pull Request that you want to comment on

lib/dialogs/add-comment.js
```diff
@@ -0,0 +1,38 @@
...
+        label: 'Search by title',
+        name: 'selectedUrl',
+        data_source: 'external',
+        min_query_length: 0,
+        optional: true,
+      },
+      {
+        type: 'text',
+        subtype: 'url',
+        label: 'Issue/pull request URL',
```
bkeepers: What about `Issue or Pull Request`?

lib/dialogs/add-comment.js
bkeepers: A nice future improvement would be to include the original author name. Something like `@wilhelmklopp said:`. This would require mapping Slack <=> GitHub usernames.

lib/dialogs/add-comment.js
wilhelmklopp: I remember a couple of discussions around this 😅 The one where I remember the outcome was about whether to use "PR" or "Pull request" and it was that we need to use "PR" because "Pull request" is too long. Slack wold not show the dialog :( https://api.slack.com/dialogs

image



There are a couple of very tight restrictions like this around dialogs unfortunately

lib/dialogs/add-comment.js
wilhelmklopp: > What do you think about just View in Slack?

I like that!

> I also wouldn't be opposed to getting rid of the <sub> to make the dialog that appears in Slack a little cleaner, but don't feel strongly about it.

Yeah, I keep coming back to re-consider the decision to have this visible to the user. There's some implementation difficulty with passing the value of the message URL along, though, so electing to keep it visible to the user for now.

> A nice future improvement would be to include the original author name. Something like @wilhelmklopp said:. This would require mapping Slack <=> GitHub usernames.

Hell yeah! We could do this today for folks who've connected their GitHub account, it just requires an additional API call. I'll open an issue for this once we've merged this PR

wilhelmklopp:
Notes for getting this into production:
We will need to go through an app directory approval process before the action actually shows up in workspaces, but the changes in this PR are backwards compatible so we can deploy them right away.

And once Slack approves and we confirm, the action will show up and our app will be ready to listen and act on those events

wilhelmklopp:
> Publishing the action now
<sub>[View in Slack](https://github.slack.com/archives/C6NQBSCJW/p1529430832000694)</sub>

wilhelmklopp:
There is a bug in the GitHub API that causes the GraphQL request that loads issues and PRs to time out. The team is looking into fixing that bug and while that investigation is ongoing this PR will be on ice.

wilhelmklopp:
> Publishing the action a second time now :crossed_fingers:
<sub>[View in Slack](https://github.slack.com/archives/C6NQBSCJW/p1529603935000633)</sub>

wilhelmklopp:
> :tada:
<sub>[View in Slack](https://github.slack.com/archives/C6NQBSCJW/p1529604387000427)</sub>

PR #607, opened 2018-05-28
WIP: Create an Issue using Message actions

mikerudge:
This integration allows users to create issues using the message_actions api. 

<a href='https://cl.ly/rsqa' target='_blank'></a>

### New ‘create-issue’ route
Route: 
 `/actions:interactive_message:actions:create-issue`

This route simply builds a dialog window object and returns to slack.  

Dialog Object
``` javascript
 const dialog = {
    callback_id: 'create-issue-dialog',
    title: 'Open new issue',
    submit_label: 'Open',
    elements: [
      {
        label: 'Repository',
        type: 'select',
        name: 'repository',
        data_source: 'external',
      },
      {
        type: 'text',
        label: 'Title',
        name: 'title',
      },
      {
        type: 'textarea',
        label: 'Write',
        value: message.text,
        name: 'body',
        placeholder: 'Leave a comment',
        hint: 'GitHub markdown syntax is supported, although you cannot preview it in Slack.',
      },
    ],
  };
```

This is the same as the `/open` object except for two major differences:

* It uses `data_source: external` to allow us to load the repos later. This makes the dialog box open as quickly as possible. 

* Instead of returning a single repo, it returns multiple repos that the user can select. 

### New Route: `/repo`

This route is called once the dialog box has opened to fill in the list of repos.  This gets all the repos that the user can access.

Returned from /repo 
``` javascript
const repos = [
  { label: "username/reponame", value: 12345 },
  { label: "username/reponame", value: 54321 }
	...
]
```

## Changes needed on the Slack app
There are minor changes needed on the slack app settings to enable the message_actions. 

### Action Settings

Simply create a new action with a callback ID of ‘create-issue’ 

action settings

### Message Menus 

This is the setting that allows for us to dynamically fill the select option with the users repo. Simply add the URL `<DOMAIN>/slack/repos`

message menus


wilhelmklopp:
Thanks for this awesome contribution! Very excited to get this merged and have this as a feature in the Slack integration! 💯 

A couple of high level points before we do a more detailed review:
- A slightly separate but naturally very similar feature is about to be merged will go live within the next day (https://github.com/integrations/slack/pull/606). I'd suggest you check that out for inspiration around structure + tests. Probably even merge that branch into this one.
- That PR also introduces a more general way for Slack to load options (not just repos), so the corresponding code introduced here would also have to use this more general approach to load repositories
> This gets all the repos that the user can access.
- I think your current implementation only lists public repositories that the user _owns_ (https://developer.github.com/v3/repos/#list-user-repositories). 
A thought: I think what we should do instead here is only load repositories that are relevant in the context of the **channel**. That seems more akin to how folks would use this feature.
Either way we probably want to use [a separate endpoint](https://developer.github.com/v3/repos/#list-your-repositories) or even GraphQL.

If we do decide to only load repos that are relevant in a specific channel, I'm happy to put some of the underlying caching in place to make that easier to implement here.

Again, really looking forward to seeing this 🚢 😄!

mikerudge:
@wilhelmklopp Thanks for the feedback! Happy to be working on such an awesome App. 

> A slightly separate but naturally very similar feature is about to be merged will go live within the next day (#606). I'd suggest you check that out for inspiration around structure + tests. Probably even merge that branch into this one.

Will definitely take a look at #606 and look to merge it into this branch, sounds like some of the problems have been solved here so that's great! 👍 

> I think what we should do instead here is only load repositories that are relevant in the context of the channel

Understand, I thought the same thing and keep going back and forth on the idea. I like that it would make the App seem 'smarter' with it showing the right repo and the right time, however, I can also see it be a cause of frustration. Our use case, for example, is, we have clients in slack who will often bring up issues in the 'wrong' channel. In those cases, not being able to create an issue might be frustrating. Likewise, seeing a huge list of repos everytime seems redundant. 

Maybe do as you said and get some feedback from the community about their use cases. 

Will look to update in the next couple of days! 👍

lib/messages/create/create-issue-attachment.js
```diff
@@ -0,0 +1,34 @@
+module.exports = (issue) => {
```
wilhelmklopp: Instead of this, could we re-use the existing message we have? https://github.com/integrations/slack/blob/master/lib/messages/issue.js

lib/create/create-issue-from-message.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+async function getUserRepos(req, res) {
+  const { slackUser } = res.locals;
+  const { channel } = req.body;
+
+  const query = `
+      query {
+        viewer {
+          repositories(first: 30, orderBy: {field: UPDATED_AT, direction: DESC}) {
```
wilhelmklopp: We may get further here by using the `repositoriesContributedTo` query with the same `orderBy` and `includeUserRepositories`: true

https://developer.github.com/v4/object/user/#repositoriesContributedTo

We also need to find a better way to get the repositories in the channel, although that would require either an additional feature in the GitHub API or some additional caching, so leave that to me to figure out 😅

lib/create/create-issue-from-message.js
```diff
@@ -0,0 +1,175 @@
...
+        });
+      } catch (error) {
+        reject(error);
+      }
+    });
+
+  const promises = repositories.nodes.map(repo => sortIntoChannel(repo));
+
+  Promise.all(promises)
+    .then(() =>
```
wilhelmklopp: We tend to use async/await directly as opposed to promises, so would love to stick to that 🙌

lib/create/create-issue-from-message.js
```diff
@@ -0,0 +1,175 @@
...
+async function createIssueFromDialog(req, res) {
+  const { slackUser } = res.locals;
+
+  const { title, body, repository } = req.body.submission;
+  const { response_url } = req.body;
+  let repos;
+  try {
+    repos = await slackUser.GitHubUser.client.repos.getById({ id: repository });
+  } catch (e) {
+    const errorMessage = JSON.parse(e.message).message;
```
wilhelmklopp: Can we be more specific here (and in the following lines that use similar logic) with respect to the error we expect, and the error we present to the user? The errors returned by the GitHub API may not translate 1:1 to something meaningful to users.

lib/dialogs/create-issue.js
```diff
@@ -0,0 +1,34 @@
+module.exports = (message, channel, team) => {
+  const linkToMessage = `https://${team.domain}.slack.com/archives/${
```
wilhelmklopp: would be awesome if we could re-use existing logic for this by extracting the common code into a module

lib/middleware/repos-attach-metadata.js
```diff
@@ -0,0 +1,22 @@
+const { SlackWorkspace, SlackUser, GitHubUser } = require('../models');
```
wilhelmklopp: I think we don't need this middleware now that we have `requireGitHubUser` middleware: https://github.com/integrations/slack/pull/606/files#diff-f02746daec04fc7c4e865ad953a4b5d2R1

lib/slack/index.js
```diff
@@ -43,28 +43,70 @@ module.exports = (robot) => {
   router.use('/actions', middleware.routeAction);
 
   // Format: '/actions:type:callback_id:action'
-  router.post(/actions:interactive_message:unfurl-\d+:unfurl-dismiss/, unfurls.showRichPreview.dismissPrompt);
-  router.post(/actions:interactive_message:unfurl-\d+:unfurl/, unfurls.showRichPreview.deliverUnfurl);
+  router.post(
```
wilhelmklopp: there seems to be some auto-formatting at play here? Could you undo these changes just to keep the diff clean? 😬

wilhelmklopp:
@mikerudge just FYI that #606 is now merged and with it the patterns and plumbing that will hopefully make your work here a little bit easier  😃

mikerudge:
@wilhelmklopp Amazing! 

Also thank you for all the feedback again! Will look into it this week hopefully!

sakulstra:
any update on this? sth i can help with? creating issues via slack would greatly improve our workflow.

lib/create/create-issue-from-message.js
alkanalkn63: True

lib/create/create-issue-from-message.js
alkanalkn63: True

lib/create/create-issue-from-message.js
alkanalkn63: True

lib/middleware/repos-attach-metadata.js
alkanalkn63: True

JoshuaCarroll:
@mikerudge Hoping you didn't abandon this - it's a great idea.

mikerudge:
Hi all! 

Soo sorry for my absence. Life!

@JoshuaCarroll Totally didn't mean to abandon this! Will try my best to try to get this across the finish line if it still makes sense to do so!

@wilhelmklopp I may need to just check in with you to see where we are at, anything I need to know etc?

It looks like quite a bit has changed since I last looked (how has it been over a year, crazy!), and after a very quick look, it seems like most (all?) the building blocks are there. 

This [issue.js](https://github.com/integrations/slack/blob/master/lib/create/issue.js) file especially looks like some big improvements have happened and looks like we have much of the work done. 

Is there something that is stopping this feature moving forward? (except my poorly timed absence)

I will probably start fresh, pull down the latest and have a play to see what I can put together with what we already have. 👍

matthiasp42:
@mikerudge would you consider finishing it?
It would be excellent for our organisation.

I would be extremely happy if this makes it into the App like it is, but I'd have one more suggestion (that I don't see here yet): 
Selecting the Assignee in the form.
I can imagine it's either really easy or a whole new story, just wanted to mention it.

JoshuaCarroll:
Yes StateBot... This is still relevant.

ashokirla:
Hi, Thanks for your contribution. We are going through platform changes to our ChatOps app and are not accepting any external contributions.  We will go through your PR and consider your suggestions if it aligns with our Product direction. We are converting your PR to a feature request and it can be tracked [here](https://github.com/integrations/slack/issues/1241).

PR #617, opened 2018-06-11
Show error message instead of 500 when github.com unfurls are disabled on workspace

wilhelmklopp:
Before:
image

Now:
image

The new error message is only shown when a user manually clicks 'Show rich preview' (where previously they would have seen the generic error message). It's not shown when the unfurl simply doesn't show up, because that could be the user-desired behavior, and a repeatedly posting warning message is quite annoying, as we've found out in the past.


wilhelmklopp:
The uncovered lines are all lines where we're re-throwing errors that don't match the error we're expecting. I Don't think it's worth it to write tests to cover those?

lib/messages/unfurls/unfurls-disabled-error.js
```diff
@@ -0,0 +1,13 @@
+const ErrorMessage = require('../flow/error-message');
+
+module.exports = class UnfurlsDisabledError extends ErrorMessage {
+  getAttachment() {
+    const helpUrl = 'https://get.slack.help/hc/en-us/articles/204399343-Share-links-in-Slack#turn-off-link-previews-for-specific-sites';
+    const attachment = {
+      ...this.getErrorMessage().attachments[0],
+      text: `Could not complete link preview, because your workspace has disabled link previews for \`github.com\` <${helpUrl}|Read more>`,
```
bkeepers: > preview, because

Unnecessary comma

wilhelmklopp:
> I assume this will show up whenever a user shares a github.com link?

Nope, only when they click the 'Show a rich preview' button 😄 🙌, so where they would already be seeing the 500 generic error message

PR #622, opened 2018-06-15
Bump html-to-mrkdwn to latest version

bkeepers:
This pulls in fixes for line breaks around headings: https://github.com/integrations/html-to-mrkdwn/pull/22

cc @devrelm


bkeepers:
@devrelm thanks for the fix. This is in production now.

Issue #507, opened 2018-04-03
Private Unfurl Feedback issue


wilhelmklopp:
Today we early-access shipped private unfurls. It's limited to only a select group of channels because the current implementation requires you to click a button every time you share a link, so it's quite annoying.
The current implementation is captured by the two gifs below:
Rich preview demo:
![show rich preview](https://user-images.githubusercontent.com/7718702/38211992-281168c0-36b4-11e8-9302-2b611590d173.gif)

Dismiss demo:
![dismiss](https://user-images.githubusercontent.com/7718702/38212000-2dc7e06e-36b4-11e8-885b-7bb4bd423a03.gif)

## What's next for private unfurls?
- [x] A proper flow for when you don't yet have your GitHub account linked
- [x] Settings
   - [x] To indicate that you always want to automatically show rich previews for this repo in this workspace
   - [x] To indicate that you always want to automatically show rich previews for all repos in this workspace?
   - [x] To say you never want to get prompted about private resources
   - [x] /github settings to change settings

## Please put feedback in this issue

If you're part of a channel where private unfurls are currently enabled, please put all relevant feedback in this issue. If you're sharing screenshots please make sure to obscure any information you don't want to be publicly visible.

bkeepers:
Might have been a fluke, but saw this:

<img width="735" alt="slack_-_github" src="https://user-images.githubusercontent.com/173/38269688-cc58251c-3746-11e8-80b9-c26bf1c669c5.png">

The message unfurled (which you can see in the top of the screenshot, but the prompt was never dismissed, and slack showed a Timeout error.

eosrei:
I'm sure many people like this, but don't want any unfurls at all. They take up far too many pixels.

bkeepers:
The prompt for unfurling a private link appears in the main channel when the link is shared in a thread.

<img width="972" alt="slack_-_github" src="https://user-images.githubusercontent.com/173/39316604-281c7452-493f-11e8-8a1b-fbd5fd62cb72.png">

wilhelmklopp:
@irvin thanks for the bug report! Did you also only see one unfurl prompt for both links?

wilhelmklopp:
@irvin I just tried to reproduce this and I don't seem to be able to. Is it possible that pull 1294 was already unfurled in the channel in the previous 30 minutes? In that case it would not have been unfurled again

irvin:
@wilhelmklopp yap, I couldn't reproduce now either. 
[...]

PR #623, opened 2018-06-15
update regex to exclude apps and topics

hiimbex:
Solves the private unfurl issue mentioned in https://github.com/integrations/slack/issues/507#issuecomment-395648098. This adds a negative lookahead to the regex to exclude apps and topics (definitely a regex noob here). Not sure how this will affect bigger picture stuff either if excluding them will break the links somewhere in the logic?

cc/ @wilhelmklopp

### DIFF
--- a/lib/github-url.js
+++ b/lib/github-url.js
@@ -4,7 +4,7 @@
 const { named } = require('named-regexp');
 
 const base = '(?:https:\/\/github.com\/)?';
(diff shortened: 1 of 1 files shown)


ashokirla:
Hi, Thanks for your contribution. We are going through platform changes to our ChatOps app and are not accepting any external contributions.  We will go through your PR and consider your suggestions if it aligns with our Product direction.

PR #626, opened 2018-06-21
Enable label picking when opening a new issue.

rapzo:
Basically it adds the possibility of choosing one label when opening an issue via slack.
It is the ground functionality but there are more in the pipeline, such as adding more labels and removing through another dialog, possibly via interactive messages.

What do you guys think?


rapzo:
Maybe i should add some description here... Hmmm.

wilhelmklopp:
> so I’d open separate PRs for those other changes
<sub>[View in Slack](https://probot-talk.slack.com/archives/C6Z70SRGD/p1529681619000697)</sub>

pipo02mix:
I would be really interesting too :)

rapzo:
Thanks! Kinda waiting on this one in order to add more funcionality, since slack selects don't allow multiple choices, some gymnastics are needed in order to add/edit labels' list.

lib/messages/create/create-issue-dialog.js
```diff
@@ -29,5 +29,19 @@ module.exports = (repository) => {
...
+  dialog.elements.push({
+    label: 'Label',
+    type: 'select',
+    name: 'label',
+    optional: true,
+    options: labels.map(({ id, name }) => ({
+      label: name,
+      value: id,
+    })),
+  });
```
wilhelmklopp: I feel like the following syntax would make this a little easier to read:

```js
if (labels) {
dialog.elements.push({
    label: 'Label',
    type: 'select',
    name: 'label',
    optional: true,
    options: labels.map(({ id, name }) => ({
      label: name,
      value: id,
    })),
  });
}
```

rapzo:
@wilhelmklopp what do you think? Messed up the response mock in order to force something else than an empty list of labels. I hope it covers it.

lib/create/issue.js
```diff
@@ -27,18 +30,23 @@ async function open(req, res) {
...
-  const { title, body } = req.body.submission;
+  const { title, body, label } = req.body.submission;
   const { repository } = req.body.submission;
 
   await slackUser.reload({ include: [GitHubUser] });
+
   const { data } = await slackUser.GitHubUser.client.repos.getById({ id: repository });
   const owner = data.owner.login;
   const repo = data.name;
+  const { data: list } = await slackUser.GitHubUser.client.issues.getLabels({ owner, repo });
```
wilhelmklopp: Can we avoid this HTTP request by passing the label name instead of the id on this line: https://github.com/integrations/slack/pull/626/files#diff-56aea2064d1a529dddab5a8b3e762507R41

I'd like to avoid additional HTTP requests as part of the dialog submission because we're on the clock 🕐 
Sorry for only bringing this up now.

lib/messages/create/create-issue-dialog.js
```diff
@@ -29,5 +29,19 @@ module.exports = (repository) => {
       },
     ],
   };
+
+  if (Array.isArray(labels)) {
```
wilhelmklopp: this is going to be true even if `labels = []` right?

Should this be something like `if (labels.length !== 0)` ?

rapzo:
Nice catches!

wilhelmklopp:
Thanks for the contribution @rapzo 💯

rapzo:
Nice! Thanks!
But i'm not done here yet! I'll be back! ;-)

PR #664, opened 2018-08-22
Replace /example with /github and reword sentence

dasevilla:


bkeepers:
Thanks @dasevilla! This will get deployed soon.

Issue #432, opened 2018-03-07
Commit comments


bkeepers:
Since commit comments are used so infrequently, we didn't implement them in the initial version. A few people have requested them, so I'm opening this just so people can track them.

This isn't a priority for us right now, but someone else is welcome to try implementing them.

cc https://github.com/integrations/slack/issues/383#issuecomment-370021318

TizocFranco:
My organization uses commit comments heavily. Is there a time frame for when this will be complete? Also, is there a way to uninstall/revert to the previous Github/Slack integration for the time being?

mattbenic:
My org also uses these commits heavily (via the API from SmartGit) for code review purposes, and we're really keen to have them back. The previous version _did_ support them.

MikeBadescu:
Code comments are important - the code provides context to the comment, helping with team communication (which is one of the purposes of GitHub and Slack). We would also like to have the code comments back.

This feature regression was not documented.

atopiler:
Our team also uses commit comments quite heavily.  It was very unexpected to see that these are no longer supported, given that they worked flawlessly in the legacy integration.

KingBrewer:
Same here. Our team would love to have this functionality back.

freemarmoset:
Would also like this functionality back.

rcaracaus:
Commenting was and still is a critical feature

jonathancounihan-sage:
+1. We use the comments to communicate with the team on specific lines on each commit. The emailing is sometimes hit and miss, but the pings in slack were easy to find and highly visible.

dominic-miglar:
+1

mfonda:
Another +1 -- our team makes heavy use of comments and would love this functionality.

> This isn't a priority for us right now

@bkeepers is this still the case, or are there any plans in the works to implement this?

shesternin0:
We need it feature too

thornycrackers:
+1

stephenbrady:
+1

ouquoi:
+1

jblanchette:
@bkeepers can you please reconsider priority; how did you classify that code review comments are "used so infrequently"?

it seems to me like this is one of the major reasons people use slack integrations and also one of the main purposes of a remote repository.

mattwynne:
@bkeepers I missed the too, and I'd be interested in having a crack at implementing it I can get some pointers.

bobgodwinx:
When is this feature going to be ready? Do we have a timeline?

[...]

PR #666, opened 2018-08-25
Support for commit comments

artnez:
As requested in #432 

Note: This enables commit comments when the `comments` setting is turned on. That means all existing users who have the `comments` setting enabled will be automatically opted into commit comments.


artnez:
Just realized this is PR #666 🎸

ayyazdaniaryan:
Cool! A sorely missed feature is coming back to life :). I think this is a great integration as is. Only part I would suggest (perhaps for a future PR) is to keep this logic and update the `Comment` object on the `edit_comment` hook. This will keep the integration clean and consistent with issue comments. Not picky at all, btw, this is great, and I cannot wait for it to be merged so we can update our slack integration and be able to read commit comments again. Thank you @artnez !

artnez:
@ayyazdaniaryan I attempted adding support for commit comment editing but couldn't find the right webhook event. 

The issue comment event has created/edited/deleted actions:
https://developer.github.com/v3/activity/events/types/#issuecommentevent

Same for pull request review comments:
https://developer.github.com/v3/activity/events/types/#pullrequestreviewcommentevent

The commit comment event only has created though:
https://developer.github.com/v3/activity/events/types/#commitcommentevent

webdif:
Our team is still using old slack integration because of the lack of this feature.
@bkeepers it would be really great if you have time to see this PR 🙏

artnez:
@bkeepers I noticed you updated to Probot 7 so I made the necessary changes on my end as well. The only thing I needed to do was rename `event` to `name` in my tests. Everything else continues to work.

bkeepers:
@artnez thanks so much for working this. This PR looks great! It's deployed to production.

---

image

PR #674, opened 2018-09-06
Manually bootstrap probot so we can load newrelic earlier

### LINKED ISSUE #370: Improve NewRelic setup
NewRelic shows this warning:

> Some modules were uninstrumented during the current time window: express, redis, bluebird. Make sure require('newrelic'); is the first line of the application's main module.

This is because Probot loads first, and then loads our `lib/index.js`. This will either require changes to Probot, or creating our own scripts to bootstrap the app.

bkeepers:
This fixes #370 by manually bootstrapping probot, which allows us to require newrelic before anything else loads.

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -1,0 +1,24 @@
+#!/usr/bin/env node
+
+require('dotenv').config();
+
+if (process.env.NEWRELIC_KEY) {
+  require('newrelic'); // eslint-disable-line global-require
+}
+
+const { findPrivateKey } = require('probot/lib/private-key');
+const createProbot = require('probot');
+const app = require('.');
+
+const probot = createProbot({
+  id: process.env.APP_ID,
+  secret: process.env.WEBHOOK_SECRET,
+  cert: findPrivateKey(),
+  port: process.env.PORT || 3000,
+  webhookPath: '/github/events',
+  webhookProxy: process.env.WEBHOOK_PROXY_URL,
+});
+
+probot.load(app);
+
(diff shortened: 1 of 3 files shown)


lib/run.js
```diff
@@ -0,0 +1,24 @@
...
+const createProbot = require('probot');
+const app = require('.');
+
+const probot = createProbot({
+  id: process.env.APP_ID,
+  secret: process.env.WEBHOOK_SECRET,
+  cert: findPrivateKey(),
+  port: process.env.PORT || 3000,
+  webhookPath: '/github/events',
+  webhookProxy: process.env.WEBHOOK_PROXY_URL,
```
wilhelmklopp: Do we need some production specific logic for this?

bkeepers: Nope. Probot will do nothing if this is blank. I don't think we're actually using smee due to the slack side, so we should probably just skip this

bkeepers:
> Do the tests we have automatically execute the "new" logic in `lib/run.js` or do we need to change something else to make that happen?

Yeah, the tests are still manually bootstrapping Probot: https://github.com/integrations/slack/blob/load-newrelic-sooner/test/integration/index.js#L12-L13

Unfortunately, none of the tests would exercise the setup. We should add new integration test that ensures that the app is bootstrapped properly (e.g. receives webhooks on the `/github/events` endpoint). I'll open an issue to track that, but don't feel like we have time to dig into it while we're trying to debug performance issues with #605

PR #675, opened 2018-09-06
Upgrade to @octokit/webhooks 5.0

bkeepers:
We're investigating some performance issues with #605, and webhook handling seems to be the culprit. Trying to upgrade @octokit/webhooks from 3.1.1 to 5.0.0 to see if we see the same performance issues against Probot 6.


bkeepers:
Confirmed that this is _not_ the culprit.

PR #676, opened 2018-09-06
avoid unfurling binary files

### LINKED ISSUE #593: Raw links to binary files spew a lot of garbage into a channel
Steps to reproduce:
Link the following two URLs in a channel that has the new GitHub For Slack integration turned on, such that GfS is doing the link expansion of the first link:
1) https://github.com/jenkinsci/emotional-jenkins-plugin/blob/master/src/main/webapp/images/sad-jenkins.png?raw=true
2) https://raw.githubusercontent.com/jenkinsci/emotional-jenkins-plugin/master/src/main/webapp/images/sad-jenkins.png

The url for 1) is obtained by doing a "right-click" / "copy image address" from the repository listing https://github.com/jenkinsci/emotional-jenkins-plugin/blob/master/src/main/webapp/images/sad-jenkins.png. We would assume it would be served up as an image, but instead it renders as though somebody threw the text content of the file into a ``` block.

screen shot 2018-05-22 at 4 53 38 pm

It'd be great if some kind of logic was used to avoid trying to render binary files as text here.
(I've also seen this trigger for links to PDFs and SVGs, to give other examples of filetypes commonly both stored in GitHub repositories and linked in Slack.)

doms:
## Background

fixes #593 

## Implementation

As @bkeepers stated here: https://github.com/integrations/slack/issues/593#issuecomment-411759744

> _We wanted to update the [blob unfurl](
https://github.com/integrations/slack/blob/master/lib/messages/blob.js) to check if the file is binary (since the [contents API](
https://developer.github.com/v3/repos/contents/#get-contents) currently doesn't have any meta data about what GitHub things the file is.)_

I used the npm package recommended in the comment and it gets the job done 👌


doms:
I understand the primary goal was to just prevent unfurling, but should we replace the content with anything? 🤔 

Currently, I went ahead and just put the text `Binary File` in the attachment since I wasn't sure if we wanted a label for it or not..

screen shot 2018-09-06 at 6 35 59 pm

lib/messages/blob.js
```diff
@@ -19,10 +20,22 @@ class Blob extends Message {
   }
 
   getCore() {
-    let lines = Buffer.from(this.blob.content, 'base64').toString().split(/\n/);
+    const buffer = Buffer.from(this.blob.content, 'base64');
+
+    let lines = buffer.toString().split(/\n/);
```
bkeepers: Could this be moved after the binary check to avoid unnecessary string manipulation?

doms:
@bkeepers 👋  

> If the blob is an image, it'd be cool to set image_url to the raw url of the image it still gets unfurled. This increases the scope of this PR, so feel free to wait to do this later.

Sounds good! 😄 I'll try to get this merged first, then start on that for the next PR. Would something like this suffice?
```js
// don't unfurl binary files
if (isBinaryFile.sync(buffer, this.blob.size)) {
  return {
    fallback: `[${this.repository.full_name}] ${title}`,
    title,
    title_link: htmlURL,
    image_url: this.blob.download_url,
  };
}
```
Granted, we would have to know for sure if the file is an image. There are other types of binary files, right?

doms:
@bkeepers bump :eyes:

bkeepers:
@doms I would expect there be a nested conditional:

1. If the file is binary…
    1. if the binary file is an image, return message with `image_url`
    1. else throw an error (for now)
1. else do the current behavior

lib/messages/blob.js
```diff
@@ -19,32 +20,51 @@ class Blob extends Message {
...
 
-    if (this.start) {
-      lines = lines.slice(
-        Number(this.start) - 1,
-        Number(this.end || this.start),
-      );
+    // only unfurl binary files that are images
+    if (isBinaryFile.sync(buffer, this.blob.size)) {
+      const imageExtensions = ['png', 'jpg', 'jpeg', 'gif'];
+      const blobExtension = this.blob.name.toLowerCase().split('.')[1];
```
doms: I wasn't too sure on how to check that the file is an image (since the contents API doesn't go into detail on _what_ the file is), so I just made a whitelist of acceptable image extensions and grabbed the file extension from the blob.

lib/messages/blob.js
```diff
@@ -19,32 +20,51 @@ class Blob extends Message {
...
+      if (imageExtensions.includes(blobExtension)) {
+        return {
+          fallback: `[${this.repository.full_name}] ${title}`,
+          title,
+          title_link: htmlURL,
+          image_url: this.blob.download_url,
+        };
+      } else { // eslint-disable-line no-else-return
+        throw new Error('File is binary');
+      }
```
doms: I opted for disabling the `no-else-return` here because it made the implementation look weird and not as readable when the `else` block was omitted (in that block of code). 

```js
if (imageExtensions.includes(blobExtension)) {
	return {
       fallback: `[${this.repository.full_name}] ${title}`,
       title,
       title_link: htmlURL,
       image_url: this.blob.download_url,
     };
}
throw new Error('File is binary');
```
Seeing that we are only throwing the error _for now_, I see this as temporary, but I can easily remove it and follow the rule.

doms:
@bkeepers ready for you to take another look whenever you get the chance 👍

gimenete:
@doms I've deployed and merged your changes. You can see the latest commits https://github.com/integrations/slack/commits/master

I'll try to merge & deploy in a different way next time so PRs like this get automatically closed.

Thanks for your PR!

doms:
@gimenete awesome, thanks!

PR #678, opened 2018-09-07
[WIP] don't show muteUnfurlPromptsUntil setting if timestamp is after now

### LINKED ISSUE #575: Check if muteUnfurlPromptsUntil timestamp is after now and don't show this setting if that is the case
https://github.com/integrations/slack/blob/7e6177f212dbdf2233eef9f7d3ef8d1b98e671e9/lib/messages/user-settings/index.js#L39

doms:
## Background

fixes #575 

## Implementation

After getting clarification from @wilhelmklopp here: https://github.com/integrations/slack/issues/575#issuecomment-419195506

The logic to **not** show the setting ended up like so:
```js
if ((this.muteUnfurlPromptsUntil < moment.unix()) && !this.muteUnfurlPromptsIndefinitely) {
  // don't show setting...
}
```
However, it wouldn't have made sense to use an if/else there, so I just did my best to negate that above snippet, and I think the solution should be to just add `moment.unix()` to the conditional:

```js
if ((this.muteUnfurlPromptsUntil > moment.unix()) || this.muteUnfurlPromptsIndefinitely) {
  // push attachment to array...
}
```

Please correct me if there is more to it! 😄

### DIFF
--- a/lib/messages/user-settings/index.js
+++ b/lib/messages/user-settings/index.js
@@ -36,7 +36,7 @@ class CombinedSettings extends Message {
       });
     }
 
-    if (this.muteUnfurlPromptsUntil || this.muteUnfurlPromptsIndefinitely) {
+    if ((this.muteUnfurlPromptsUntil > moment.unix()) || this.muteUnfurlPromptsIndefinitely) {
       attachments.push({
         ...super.getBaseMessage(),
         title: 'Muted prompts to show rich preview',


doms:
I'm not too sure on how to fix the tests...

Do I need to **update** the failing snapshots? I did some digging through the [jest snapshot testing docs](https://jestjs.io/docs/en/snapshot-testing.html) because no matter what I did, the tests didn't pass, then I realized, maybe the behavior is right, but the snapshots need to be updated! (since its expecting the _old_ value in the `*-test.js.snap` file)

I'm not sure if this could be remedied with just changing the logic of the tests, or if updating the snapshots is a must. 🤔

These are the tests (and their relevant snapshots) that are failing:

https://github.com/integrations/slack/blob/0cfaabaa04fe5db6a4d8c33d5d71ad1760a2f32a/test/integration/user-settings.test.js#L135-L152
https://github.com/integrations/slack/blob/0cfaabaa04fe5db6a4d8c33d5d71ad1760a2f32a/test/integration/__snapshots__/user-settings.test.js.snap#L164-L181

---

https://github.com/integrations/slack/blob/0cfaabaa04fe5db6a4d8c33d5d71ad1760a2f32a/test/messages/user-settings/index.test.js#L51-L60
https://github.com/integrations/slack/blob/0cfaabaa04fe5db6a4d8c33d5d71ad1760a2f32a/test/messages/user-settings/__snapshots__/index.test.js.snap#L76-L93

wilhelmklopp:
Just glanced at it briefly. Will take a more in depth look later.

I believe we want the existing snapshot tests to still be green, but we also want a new snapshot test where `muteUnfurlPromptsUntil` is set to some value that is less than what we mock to be the current time (what moment.unix() would return) and where `muteUnfurlPromptsIndefinitely` is not set. In that case we'd want the user to see the `"text": "You do not have any settings configured.",` message.

I _also_ think there is some weirdness in this that I'm not fully understanding, so hang in there :D

lib/messages/user-settings/index.js
```diff
@@ -36,7 +36,7 @@ class CombinedSettings extends Message {
       });
     }
 
-    if (this.muteUnfurlPromptsUntil || this.muteUnfurlPromptsIndefinitely) {
+    if ((this.muteUnfurlPromptsUntil > moment.unix()) || this.muteUnfurlPromptsIndefinitely) {
```
wilhelmklopp: Should this comparison sign be the other way around based on our discussion here?

https://github.com/integrations/slack/issues/575#issuecomment-419236694

wilhelmklopp: Ah, no. I think I'm wrong 🤔

wilhelmklopp:
@doms I think I've figured out what's going wrong.

Instead of this:
```
moment.unix()
```

You'll want this
```js
moment().unix()
```

With that change, the tests should pass and we just need to add two tests:
- one test in `test/messages/user-settings/index.test.js`
- one test in `test/integration/user-settings.test.js`

Feel free to copy from the tests that are already there 🎆

lib/messages/user-settings/index.js
doms: > _Should this comparison sign be the other way around based on our discussion here?_

I thought the same thing initially! But then I realized that entire conversation was us deciding on what we should do to **not** show the setting. Sooo for the implementation, I had to _negate_ our discussion logic (it confused me for quite a bit 😆 )

> Ah, no. I think I'm wrong

You're not wrong, per se, it's just confusing perspective!

doms:
@wilhelmklopp 

> Instead of this:
> 
> ```
> moment.unix()
> ```
> You'll want this
> 
> ```js
> moment().unix()
> ```

So I just needed one more set of parenthesis to solve all my issues... 🤔 

Thanks for investigating for me! (I'm still getting used to the codebase and must have overlooked this simple oversight)

I'll add the tests and push it up shortly after! 👍

wilhelmklopp:
@doms are you happy for me to take another look + merge if everything looks good?

doms:
@wilhelmklopp yea! Sorry, I should have mentioned you/requested a review when I was ready for you to take another look 😄

doms:
@wilhelmklopp any updates? 👀

test/integration/user-settings.test.js
```diff
@@ -151,6 +151,25 @@ describe('Integration: User settings', async () => {
         });
     });
 
+    test('with prompts muted temporarily doesn\'t see setting showing time remaining', async () => {
+      Date.now = jest.fn(() => new Date(Date.UTC(2018, 4, 4)).valueOf());
+      await slackUser.update({
+        settings: {
+          muteUnfurlPromptsUntil: 1825398771,
```
wilhelmklopp: I believe this value is causing the `10 years` in the snapshots.

image

For the snapshot to give us the desired output, i.e. test the new change in this PR, I believe we want this unix timestamp for `muteUnfurlPromptsUntil` to be `before` the `new Date(Date.UTC(2018, 4, 4)` value.

test/integration/__snapshots__/user-settings.test.js.snap
```diff
@@ -161,6 +161,26 @@ Object {
...
+      "actions": Array [
+        Object {
+          "name": "unmute-prompts",
+          "text": "Unmute",
+          "type": "button",
+        },
+      ],
+      "callback_id": "unfurl-settings",
+      "color": "#24292f",
+      "text": "Prompts to show a rich preview are muted for 10 years",
```
wilhelmklopp: > for 10 years

😱

test/messages/user-settings/index.test.js
```diff
@@ -58,4 +58,15 @@ describe('CombinedSettings rendering', () => {
     }).getAttachments();
     expect({ attachments }).toMatchSnapshot();
   });
+
+  test('doesn\'t work if temporarily muted prompts is after current time', async () => {
```
wilhelmklopp: Could we update this test name to be something more like "doesn't show if temporary mute is no longer active"

test/integration/user-settings.test.js
doms: > I believe this value is causing the 10 years in the snapshots.

oops 😆 Thanks for clearing this up, I will address this! 👍

doms:
@wilhelmklopp Ready for another look when you get the chance 👀

PR #684, opened 2018-09-13
Show pull request reviewers

### LINKED ISSUE #655: In the message list the `Reviewers` under `Assignees` field when PR is opened
This is a new feature request, namely to have the list of `Reviewers` in addition and similar to how `Assignees` are shown here:
image

gimenete:
Adds the usernames of the requested reviewers and the ones that have already reviewed a pull request.

Closes #655


lib/messages/pull-request.js
```diff
@@ -48,6 +59,10 @@ class PullRequest extends AbstractIssue {
         title: 'Comments',
         value: this.pullRequest.comments,
       },
+      {
+        title: 'Reviewers',
+        value: Array.from(reviewers).join(', '),
```
wilhelmklopp: Are there emojis that we can put here that map well to the: commented, requested, approved, requested_changes states? It's a long shot but maybe possible?

test/integration/notifications.test.js
```diff
@@ -198,6 +198,7 @@ describe('Integration: notifications', () => {
           Accept: 'application/vnd.github.html+json',
         },
       }).get('/repos/github-slack/app/issues/31').reply(200, fixtures.issue);
+      nock('https://api.github.com').get('/repos/github-slack/app/pulls/31/reviews').reply(200);
```
wilhelmklopp: I think we want to always return a fixture with these stubbed out HTTP requests, similar to the line above. `fixtures.review` could look something like your example from `test/messages/pull-request.test.js`:
```js
[
        { state: 'CHANGES_REQUESTED', user: { login: 'user2' } },
        { state: 'APPROVED', user: { login: 'user2' } },
      ],
```

lib/messages/pull-request.js
gimenete: That's a great idea! As we discussed offline I think we can do that in another issue because a user can do multiple reviews and we only want not the last one but the last one that is meaningful, because the last one can be a comment and there can be a previous one that is approved or request_changes,... so I think we can do that in another issue and keep this one simple. But I'm 100% for implementing it!

Issue #642, opened 2018-07-16
New issues opened via `/github open` don't use template


jamesroseman:
When using `/github open owner/repo` to file a new ticket in my repository, I noticed that the prompt that showed up did not use the template set within the repository.

<img width="534" alt="screenshot 2018-07-16 15 17 55" src="https://user-images.githubusercontent.com/1159190/42778541-6cfe2370-890b-11e8-9d86-6172e1f867cf.png">

This is what it looks like to open a new issue in the repository:

<img width="791" alt="screenshot 2018-07-16 15 18 36" src="https://user-images.githubusercontent.com/1159190/42778572-8b0ab130-890b-11e8-93e8-a938e4e3c6c4.png">

We use this Slack app to help our non-tech file tickets. We're a tiny startup, so this process helps us a lot.

hedqvist:
Experiencing same issue here, would be great if the template could be used!

andymckay:
There's a PR in issue #686 with a question, what should we do when there are multiple templates? The proposed solution allows the template name to be passed in as a parameter.

twebber90:
Looks like this issue was closed via a merged PR (a while ago) but I don't see any documentation for this functionality.. Also looked through the code in this PR to attempt to glean how it functions and I haven't been able to get it working by passing in additional parameters.

Would love to see this functionality in the integration 👍

twebber90:
Sorta disregard ☝️ 

Figured out that you need to have issue templates in the repo that you are adding an issue to (we use an org level `.github` repo to standardize our templates across repos).

IMO it would be helpful to have that requirement documented somewhere. Also, support for org level templates would be cool :+1:

Awesome that we got this working regardless 🙌

PR #686, opened 2018-09-14
Use issue template

gimenete:
Closes https://github.com/integrations/slack/issues/642


gimenete:
@wilhelmklopp I need to add tests and I have a question. What should we do if there is more than one issue template? How can we give the user the ability to choose one? We can add a new param to the slash command and found the closest file name e.g. `/github open repo template-name` or we can do something in the slack UI. Ideas?

andymckay:
Here's a blog post about it: https://blog.github.com/2018-01-25-multiple-issue-and-pull-request-templates/

PR #693, opened 2018-09-18
Show teams in reviewers attachment

gimenete:


gimenete:
@bkeepers I've just pushed a commit to implement your suggestion. Take a look! I thought the text in the attachments was going to work as a mention to slack users, but it's not the case, so ignore my previous comment.

PR #700, opened 2018-09-26
We are working on slack...

andymckay:
Thanks to @gimenete we've been doing some slack work. Whilst this might be temporary, I think this sends a more accurate signal over the next month or so.


ahmad380360:
Ok

PR #701, opened 2018-09-28
Fix the broken license link

philipjohn:
There's no `.md` extension on the actual file in the repo, so this link 404s. This change fixes that by fixing the reference.


bkeepers:
Thanks!

PR #706, opened 2018-10-04
#574 Fix release empty title

thrawny:
Fixes #574
Fall back to tag name when the release name is not set.


thrawny:
Anyone taking a look at this?

agoodno:
@thrawny I just hit the same issue this morning. Did you just close this because it wasn't getting any attention? Does it fix the problem? I'm wondering how we can move this forward.

PR #709, opened 2018-10-09
Feature/organization subscriptions

gimenete:
Based on what @GertSallaerts started here https://github.com/integrations/slack/pull/602

I've continued mainly with a change in the database. There's now a new column to know what thing we are subscribing to: `repo` or `account` at this moment. This way we don't need to guess what we are subscribed to doing multiple requests and seeing which one does not return 404.

Additionally I've changed a few things:
- how the information is rendered in `subscribe list`
- the way an installation is found for an account
- the way subscriptions related to an event are queried
- tests


lib/messages/flow/help.js
```diff
@@ -8,7 +8,15 @@ const commands = [
   },
   {
     usage: 'unsubscribe owner/repository',
-    desc: 'Unsubscribe to notifications for a repository',
+    desc: 'Unsubscribe from notifications for a repository',
+  },
+  {
+    usage: 'subscribe owner',
+    desc: 'Subscribe to notifications for an organization',
```
wilhelmklopp: Should we make explicit here that you "subscribe to notifications for all repositories in an organization"? Since there are a bunch of other org level events around teams, members, and so on that won't show up in Slack?

lib/messages/subscription-list.js
```diff
@@ -18,28 +19,40 @@ module.exports = class SubscriptionList extends Message {
       attachments: [{
         ...this.getBaseMessage(),
         fallback: `${prefix} ${this.repositories.length} repositor${this.repositories.length === 1 ? 'y' : 'ies'}`,
+        title: `${prefix} these repositories`,
+        text: this.repositories.length > 0 ? this.repositoriesToString().join('\n') : 'no repositories',
+      }, {
+        ...this.getBaseMessage(),
+        fallback: `${prefix} ${this.accounts.length} account${this.accounts.length === 1 ? '' : 's'}`,
+        title: `${prefix} these accounts`,
+        text: this.accounts.length > 0 ? this.accountsToString().join('\n') : 'no accounts',
```
wilhelmklopp: Looks like this currently works out to the following in case of no repos and no accounts: 
image

Could we change this so that if a channel is not subscribed to any repositories, only the list for accounts is shown? And similarly if a channel is not subscribed to any accounts, only the list of repositories is shown?

If a channel is not subscribed to any repositories and also not subscribed to any accounts can we say something like: "#channel is not subscribed to any repositories or accounts"?

lib/github/get-accounts.js
```diff
@@ -0,0 +1,19 @@
+const logger = require('../logger');
+
+module.exports = async (subscriptions, robot) => {
+  const accounts = await Promise.all(subscriptions.map(async (subscription) => {
+    const github = await robot.auth(subscription.Installation.githubId);
+    try {
+      const account = await github.users.getById({ id: subscription.githubId });
+      return account.data;
+    } catch (err) {
+      logger.error({ err, accountId: subscription.githubId }, 'Could not find account for subscription');
```
wilhelmklopp: I don't believe we would ever get a 404 back from this because all GitHub accounts are public. So even with our installation token, we'll always get an account back here.

lib/models/installation.js
```diff
@@ -28,21 +28,31 @@ module.exports = (sequelize, DataTypes) => {
...
+          repo,
+        }
+        : {
+          method: 'GET',
+          url: '/users/:username/installation',
+          headers: {
+            accept: 'application/vnd.github.machine-man-preview+json',
+          },
+          username: owner,
+        };
```
wilhelmklopp: GitHub Apps Question: If someone has installed the "Slack" GitHub App on a single repository, then will making an API request  to `/users/:username/installation` tell us about whether it's installed on the whole org or just on that one repo?

lib/models/installation.js
```diff
@@ -28,21 +28,31 @@ module.exports = (sequelize, DataTypes) => {
...
+      const { data } = await github.app.request(options);
       // Reuse install, which already does a findOrCreate
       return Installation.install(data);
     },
 
     async install(payload) {
+      // const ownerId = payload.type === 'Or'
```
wilhelmklopp: I don't fully understand this comment 😅

lib/commands/subscribe.js
```diff
@@ -15,15 +15,23 @@ module.exports = async (req, res) => {
...
+    } else {
+      from = (await gitHubUser.client.repos.get({
+        owner: resource.owner,
+        repo: resource.repo,
+      })).data;
+    }
   } catch (err) {
-    req.log.debug({ err }, 'Could not find repository');
+    req.log.debug({ err }, `Could not find ${resource.type}`);
     return command.respond(new NotFound(command.args[0]).toJSON());
```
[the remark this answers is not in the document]
gimenete: Yes, it has been updated

lib/github/get-accounts.js
gimenete: What if an account is deleted?

lib/models/installation.js
gimenete: hahaha, neither do I

lib/github/get-accounts.js
wilhelmklopp: 🙈 Yes, that's a very good point hahaha 😅 

I think I mentioned this because we often use the get-repo equivalent as a way to find out whether our current access token has access to a repository, but that doesn't apply in this case. My bad! 😆

gimenete: No problem! 😉

PR #711, opened 2018-10-11
Replace deprecated oauth.token method

Sytten:
__Why__
As you know workspace applications are deprecated. Hence, it is not possible for new contributors to create one and use it, we have to rely on a _normal_ application.

__How__
For _normal_ applications, the `oauth.token` method does not work. So I switched it to the recommended endpoint. I also had to change the scope since `chat:write` does not exists for _normal_ applications.

Since the official github app is still a workspace app, I will need help from a dev at github to test it since the endpoints don't return the same things for workspace apps.


tebriel:
Hi! Thanks for your contribution (and sorry that it's been more than a year since we got back to you). We have made this change internally (but are not able to yet release it). I'm closing it, but I'm working on pushing a commit back with those changes to remove the oauth.token use and switch to oauth.access!

Sytten:
Yeah this is kind of a blocker for any new contributor since we can't create new app tokens.

tebriel:
@Sytten understood, thanks for that feedback, we need to do an overhaul on how to stand up the app for development to make this process easier.

PR #730, opened 2018-11-03
Fix subscription query

gimenete:


gimenete:
@wilhelmklopp yeah, I'll do a test in a follow up PR 👍

lib/models/subscription.js
```diff
@@ -143,12 +143,10 @@ module.exports = (sequelize, DataTypes) => {
...
     },
 
-    async lookupAll(...githubIds) {
+    async lookupAll(query) {
       return this.findAll({
         where: {
-          githubId: {
-            [Op.or]: githubIds,
-          },
+          [Op.or]: query,
```
wilhelmklopp: non-blocking nit: Instead of `query`, should we call this `conditions`?

PR #745, opened 2018-12-14
Remove crypto key reuse

gimenete:


tcbyrd:
@gimenete also updated the OP to reference https://github.com/github/ecosystem-primitives/issues/91 instead of 94

lib/index.js
```diff
@@ -23,7 +28,11 @@ module.exports = (robot) => {
   setupApi(robot);
 
   app.use(session({
-    keys: [process.env.GITHUB_CLIENT_SECRET],
+    // See https://github.com/expressjs/cookie-session#keys
+    // Remove GITHUB_CLIENT_SECRET in a future PR
+    // Since GITHUB_CLIENT_SECRET is not keys[0] it is not used for signing new cookies,
+    // just for validating old ones
+    keys: [SESSION_SECRET, process.env.GITHUB_CLIENT_SECRET],
```
wilhelmklopp: Ah, this is great!

gimenete:
@wilhelmklopp you are right. I've just added the env variable to both: staging and production. So I'll deploy this

PR #748, opened 2018-12-18
Webhook not await

gimenete:


lib/activity/processor.js
```diff
@@ -0,0 +1,6 @@
+module.exports = (logger) => {
```
wilhelmklopp: I think a comment here to this PR or a brief explanation of what this `processor` does would be awesome 💯

philipbremer: A bit of documentation on your thought process would be helpful here and in the tests.

philipbremer:
@gimenete - to be explicit, will you add your rollout plan in the PR description please?

PR #750, opened 2018-12-19
Org unit tests

gimenete:
Closes https://github.com/github/ecosystem-primitives/issues/97

This is implementing additional tests for authentication. It tests that from a list of commands, all of them require to prompt the user for GitHub authentication if none is provided.

### DIFF
--- a/test/integration/__snapshots__/signin.test.js.snap
+++ b/test/integration/__snapshots__/signin.test.js.snap
@@ -15,6 +15,96 @@ Object {
 }
 `;
 
+exports[`Integration: signin unauthenticated user is prompted to authenticate for slash command "close https://github.com/owner/repo/issues/123" 1`] = `
+Object {
+  "attachments": Array [
+    Object {
+      "color": "#24292f",
+      "mrkdwn_in": Array [
+        "text",
+      ],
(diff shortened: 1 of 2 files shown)


tcbyrd:
@gimenete Still learning how this works a bit.

> prompt the user for GitHub authentication **if none is provided.**

Where do we determine if the authentication has been provided? I see that this tests the full flow, but I'm not clear where in the test we decide whether or not the user needs to go through this flow.

gimenete:
@tcbyrd in order to simulate an authenticated user tests create a couple of records in the db like this:

```js
    let githubUser;
    let slackUser;
    beforeEach(async () => {
      githubUser = await GitHubUser.create({
        id: 1,
        accessToken: 'secret',
      });

      slackUser = await SlackUser.create({
        slackId: 'U88HS', // same as in link_shared.js
        slackWorkspaceId: workspace.id,
        githubId: githubUser.id,
      });
    });
```

Then, a Slack event is simulated:

```
js
await request(probot.server).post('/slack/events').send(fixtures.slack.link_shared())
        .expect(200);
```

The payload contains the slack id of the user. In this case `U88HS`. If the slack id doesn't correspond to a SlackUser or there's no GitHubUser that matches that SlackUser, then the user is not authenticated against GitHub.

In this PR all tests start with an empty db, so there's no possible user that could be authenticated.

PR #754, opened 2018-12-26
Soft delete subscriptions

gimenete:
Closes https://github.com/github/ecosystem-primitives/issues/103

A way of soft-deleting subscriptions could have been adding a `disabled` column or something like that. But I thought it was better and simpler to just create a new table storing that historic information for a couple of reasons:

- Security: we will never forget to specify `disabled: false` when querying subscriptions
- Performance: the table of deleted subscriptions will grow and grow, but there won't be a performance impact when querying active subscriptions.

I've decided to use the hooks feature from Sequelize so the implementation is as transparent and reusable as possible and I've added tests in all those cases I've found that subscriptions are deleted.


tcbyrd:
Like the implementation here and using hooks. The way this is configured now, if a user re-subscribes, they would be starting over from scratch, but all their previous settings would be stored in the `DeletedSubscriptions` table under a different ID, correct? I wonder if we should a prompt for the user to restore settings from a previously deleted subscription?

gimenete:
Hmmm restoring the settings would be a new feature and I'm not sure if it's worth it. Here the idea is just to keep a log of deleted subscriptions so if we need to investigate an incident we can match the deliveries with subscriptions, even if they were deleted at some point.

db/migrations/20181221164327-create-deleted-subscription.js
```diff
@@ -0,0 +1,48 @@
+module.exports = {
+  up: (queryInterface, Sequelize) => queryInterface.createTable('DeletedSubscriptions', {
+    id: {
+      allowNull: false,
+      autoIncrement: true,
+      primaryKey: true,
+      type: Sequelize.BIGINT,
+    },
```
wilhelmklopp: should we also have a column that represents the `id` from the `Subscriptions` table? Since the primary id in this new table will likely have a different value

test/integration/notifications.test.js
```diff
@@ -494,7 +495,7 @@ describe('Integration: notifications', () => {
       });
     });
 
-    test('post prompt to re-subscribe after user loses access to repo', async () => {
+    test.only('post prompt to re-subscribe after user loses access to repo', async () => {
```
wilhelmklopp: rogue `only` -- this is probably causing the drop in test coverage?

db/migrations/20181221164327-create-deleted-subscription.js
gimenete: hmmm, what do you mean? I don't think we care about the original subscription id. But if we care I think the easiest solution would be to make the id the same and drop the `autoIncrement` in the `DeletedSubscriptions` table. What do you think?

test/integration/notifications.test.js
gimenete: Whooops. Thanks for spotting that!

db/migrations/20181221164327-create-deleted-subscription.js
wilhelmklopp: > hmmm, what do you mean? I don't think we care about the original subscription id.

I was thinking from the perspective of needing to reconcile a deleted subscription with perhaps a log message somewhere that includes just the subscription id (but nothing else).

> But if we care I think the easiest solution would be to make the id the same and drop the autoIncrement in the DeletedSubscriptions table. What do you think?

Agree, this seems like the better solution 👍

gimenete:
I'm not going to merge this yet until the PR about adding names to the db is merged, because I'll need to add those new columns in the deleted subscriptions too.

lib/activity/router.js
```diff
@@ -57,13 +57,13 @@ module.exports = function route(callback) {
...
                 githubId: subscription.githubId,
+                githubName: subscription.githubName,
                 workspaceId: subscription.SlackWorkspace.slackId,
               },
             }, 'User lost access to resource. Deleting subscription.');
 
             await Promise.all([
-              // @todo: deactive this subscription instead of deleting the db record
-              await subscription.destroy(),
+              await subscription.destroyWithReason('lost access'),
```
wilhelmklopp: could we update this to `subscription creator lost access`?

```suggestion
              await subscription.destroyWithReason('subscription creator lost access'),
```

lib/slack/uninstall.js
```diff
@@ -14,7 +14,7 @@ module.exports = async (req, res) => {
   });
 
   if (subscriptions) {
-    await Promise.all(subscriptions.map(subscription => subscription.destroy()));
+    await Promise.all(subscriptions.map(subscription => subscription.destroyWithReason('uninstall')));
```
wilhelmklopp:
```suggestion
    await Promise.all(subscriptions.map(subscription => subscription.destroyWithReason('slack app uninstalled')));
```

db/migrations/20181221164327-create-deleted-subscription.js
```diff
@@ -0,0 +1,51 @@
...
+      allowNull: false,
+      type: Sequelize.DATE,
+    },
+    reason: {
+      type: Sequelize.STRING,
+      allowNull: true,
+    },
+  }),
+  down: queryInterface => queryInterface.dropTable('DeletedSubscriptions'),
+};
```
wilhelmklopp: ⚠️ scope creep alert ⚠️ 

Do we have any way of making sure that the columns stay in sync even when someone is making changes to the Subscriptions table?
Perhaps a test that asserts that the columns are the same (accounting for `reason`, `deleatedAt`, etc.)

db/migrations/20181221164327-create-deleted-subscription.js
```diff
@@ -0,0 +1,51 @@
...
+      type: Sequelize.JSON,
+    },
+    createdAt: {
+      allowNull: false,
+      type: Sequelize.DATE,
+    },
+    updatedAt: {
+      allowNull: false,
+      type: Sequelize.DATE,
+    },
```
wilhelmklopp: do we want a `deletedAt` column? (assuming that the values from `createdAt` and `updatedAt` are copied directly from the real subscription)

lib/models/subscription.js
```diff
@@ -133,6 +139,17 @@ module.exports = (sequelize, DataTypes) => {
       this.changed('settings', true);
     },
 
+    async updateGithubName(githubName) {
+      if (this.githubName !== githubName) {
+        await this.update({ githubName });
+      }
+    },
```
wilhelmklopp: 💡 I think it might be good to protect against accidentally updating to `null` here. if the passed in `githubName` is `null` (passed by accident from the calling function), then we should probably abort and log a warning?

lib/models/subscription.js
```diff
@@ -173,7 +190,12 @@ module.exports = (sequelize, DataTypes) => {
     },
 
     async unsubscribe(githubId, channelId, slackWorkspaceId) {
-      await this.destroy({ where: { githubId, channelId, slackWorkspaceId } });
+      const where = { githubId, channelId, slackWorkspaceId };
+      const subscriptions = await this.findAll({ where });
+      await this.destroy({ where, individualHooks: true });
+
+      const ids = subscriptions.map(subscription => subscription.id);
+      await sequelize.model('DeletedSubscription').update({ reason: 'unsubscribe' }, { where: { id: ids } });
```
wilhelmklopp: it'd be cool if we could pass a reason + multiple subscriptions into `destroyWithReason`, otherwise this could be a bit race-conditiony?

lib/models/subscription.js
```diff
@@ -197,7 +219,10 @@ module.exports = (sequelize, DataTypes) => {
       });
       await this.destroy({
         where: { creatorId },
+        individualHooks: true,
       });
+      const ids = subscriptions.map(subscription => subscription.id);
+      await sequelize.model('DeletedSubscription').update({ reason: 'signout' }, { where: { id: ids } });
```
wilhelmklopp: same question/concern here as above

db/migrations/20181221164327-create-deleted-subscription.js
gimenete: There's at least one test that validates all the fields of the subscription.

https://github.com/integrations/slack/pull/754/files#diff-d662e6c04a8d262a012ac1efa9a6f9f7R103

The `reason` column is tested in many (all?) tests regarding deleted subscriptions.

I can be more explicit about other tests where I just test the count() of deletedsubscriptions, but since all the deleted subscriptions are created the same way (with `create(subscription.dataValues)` I didn't see it as necessary.

db/migrations/20181221164327-create-deleted-subscription.js
gimenete: Oh, that's a good idea. I'll add that.

gimenete: Done :)

lib/models/subscription.js
gimenete: I added a conditional. I don't think we have to abort because if some webhook changes or a new webhook event has a different _shape_ we might be aborting the deletion for not a good reason.

lib/models/subscription.js
gimenete: I'm not doing that because I wanted to keep the bulk delete. I think this is very covered by tests so I would keep it like it is.

PR #755, opened 2018-12-27
Test cache access

gimenete:
Closes https://github.com/github/ecosystem-primitives/issues/106

In order to test what we put inside Redis and whether or not we are accesing it, I've added spies to the `cache` module. Similar to what we do with `nock`. I'm testing which values are accessed (`get`) and which ones are `set`. I could not test all the calls with `expect(spy).toMatchSnapshot()` because execution order is not guaranteed when we are using `async` functions.

For example both, the `matchMetaDataStatetoIssueMessage` function and the `comments` function in `lib/activity.js` are called for the `issue_comment.created` and `issue_comment.edited` events. But the execution order of those functions changes, even when running multiple times the same test. In the future it would be great to be able to just test one of those functions directly, but right now there's no easy way to do that, unless we do an important refactoring.

I found also that since two functions are being executed for the same event type, we are effectively executing the `route()` function twice. It would be a lot better to be able to specify like a middleware to probot since we are using the `route()` function for all types of events.

### DIFF
--- a/test/integration/__snapshots__/actions.test.js.snap
+++ b/test/integration/__snapshots__/actions.test.js.snap
@@ -1,10 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Actions An unknown callback_id returns a 500 1`] = `[MockFunction cache.get]`;
+exports[`Actions An unknown callback_id returns a 500 1`] = `Set {}`;
 
-exports[`Actions An unknown callback_id returns a 500 2`] = `[MockFunction cache.set]`;
-
-exports[`Actions An unknown callback_id returns a 500 3`] = `[MockFunction cache.fetch]`;
+exports[`Actions An unknown callback_id returns a 500 2`] = `Map {}`;
 
 exports[`Actions When an action's name is "cancel", then it is immediately deleted 1`] = `
 Object {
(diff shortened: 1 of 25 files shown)


gimenete:
I've found that there is at least one test that produces different snapshots depending on how it is run. If it is run isolated or if the whole `notifications.test.js` file is executed. The test name is `comments are updated when edited`.

After a lot of debugging I've found that in `lib/activity.js` we are subscribing twice to `issue_comment.edited`. The first time with the function called `matchMetaDataStatetoIssueMessage` and the second time for the `comments` function. It seems that probot is not using the standard Node.js EventEmitter that guarantees the order in which listeners are executed. Instead it is using an external package: `promise-events` that does not seem to guarantee the order in which listeners are executed. I'm talking about the sync stuff until either `matchMetaDataStatetoIssueMessage` or `comments` are called. Because of that, cache keys are sometimes generated in different order and CI fails.

wilhelmklopp:
> For example both, the matchMetaDataStatetoIssueMessage function and the comments function in lib/activity.js are called for the issue_comment.created and issue_comment.edited events. But the execution order of those functions changes, even when running multiple times the same test. In the future it would be great to be able to just test one of those functions directly, but right now there's no easy way to do that, unless we do an important refactoring.

> I found also that since two functions are being executed for the same event type, we are effectively executing the route() function twice. It would be a lot better to be able to specify like a middleware to probot since we are using the route() function for all types of events.

Really great find! Should we open a `chore` issue for this on this repo?

gimenete:
> Should we open a chore issue for this on this repo?

Good idea. I'll do that.

> Is the cache key always included in Map {}? If so, should we just store that snapshot and not the Set of keys separately?

🤔 Yeah, you are right. I'll simplify it.

gimenete:
@wilhelmklopp I think it's ok to test both the `get` access and the `set` access. Could it be that you `get` something and you don't `set` it, or you `set` it but you don't `get` it.

If in general we always `set` if a `get` doesn't succeed, but in a particular case by mistake we don't do it, this would be a way for catching the mistake.

PR #758, opened 2019-01-04
Store github names for repos, orgs and users

gimenete:
Closes https://github.com/github/ecosystem-primitives/issues/108

There's no easy way to get the names of repos, orgs and users given their github id, and that's something we need to do if we need to investigate an issue. So, even though names can change and information can be out of date, it is worth it to store it and only recalculate those that may have changed in case of an investigation.

For subscriptions I'm storing the name when the subscription is created and if we detect the name has changed during a webhook processing.

For github users, I'm storing and updating if needed when the user signs in.


lib/github/oauth.js
```diff
@@ -73,14 +73,15 @@ module.exports = (robot) => {
       token: accessToken,
     });
     const user = (await github.users.get({})).data;
+    const defaults = { accessToken, login: user.login };
 
     // store github user id and connect to slack user id
     const [gitHubUser, created] = await GitHubUser.findOrCreate({
       where: { id: user.id },
-      defaults: { accessToken },
+      defaults,
```
cktricky: Hey @gimenete 👋 ,

Just wanted to make sure my assumption is correct here and that both values inside of `default` have to match in order to return a user object. Either way, should be good from a security perspective as this is the result of authenticating against GitHub to retrieve user details.

gimenete: Only the fields in the `where` have to match in order to find the user. If it's not found, a user is created with the attributes resulting of merging the `where` and the `default` ones. In the next sentence there's an `if` to update the user with the values in `defaults` if the user was found (not created).

For additional information, the `user` object comes from the GitHub API so we know that the `id` is unique and never changes.

lib/activity/router.js
```diff
@@ -76,6 +77,13 @@ module.exports = function route(callback) {
         // Delay GitHub API calls to avoid replication lag
         context.github.hook.before('request', avoidReplicationLag());
 
+        const githubName = subscription.type === 'account'
+          ? context.payload.repository.owner.login
+          : context.payload.repository.full_name;
+        if (subscription.githubName !== githubName) {
+          await subscription.update({ githubName });
+        }
+
```
wilhelmklopp: I wonder if it makes sense to have this logic in the `Subscription` model as a method like `updateName`?

lib/models/github-user.js
```diff
@@ -20,6 +20,8 @@ module.exports = (sequelize, DataTypes) => {
       type: DataTypes.STRING,
       allowNull: false,
     }),
+
```
wilhelmklopp: Should we add a note here saying for what reason we store this value? Or mention that this isn't used in the application and only intended for debugging?

lib/models/subscription.js
```diff
@@ -39,6 +39,7 @@ module.exports = (sequelize, DataTypes) => {
 
   const Subscription = sequelize.define('Subscription', {
     githubId: DataTypes.BIGINT,
+    githubName: DataTypes.STRING,
```
wilhelmklopp: same question as above 😬 -- Could we add a comment about what it's used for?

db/migrations/20190104111051-add-name-columns.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('SlackUsers', 'githubName', {
+      type: Sequelize.STRING,
+    });
```
wilhelmklopp: Is this `SlackUsers` column used for anything? I can't seem to find it in the rest of the changes.

lib/activity/router.js
gimenete: Sounds good. Done!

lib/models/github-user.js
gimenete: Makes sense. Done

lib/models/subscription.js
gimenete: Done here as well 👍

db/migrations/20190104111051-add-name-columns.js
gimenete: You are right. Deleting this.

lib/github/oauth.js
cktricky: Thanks @gimenete - I was just trying to figure out if there would be any ramifications if the one piece of information under user control (`user.login` - since they can change their login information) would be impacted. For example, a scenario where:

* 1 - The `acme` org changes their login to `theacme`
* 2 - A user takes advantage of this login name change and takes the old org name `acme`

I suppose it wouldn't matter as we **first** perform the lookup by the `id` which doesn't change as you'd mentioned.

gimenete: We are storing the `login` just for debugging purposes and we are always querying by `id` and primary keys are `ids` too. Just to make that clear, as @wilhelmklopp suggested it, I've also added the following comment to the new columns that store login/names:

> Storing only for debugging purposes. Do not use for querying data since it might change

cktricky: Perfect 👌 ! Thank you @gimenete

PR #765, opened 2019-01-17
Scope queries and cache keys by workspace id

gimenete:
Closes https://github.com/github/ecosystem-primitives/issues/98

Channel id uniqueness is not guaranteed, even though in practice they are unique. This PR enforces always scoping queries and cache keys by workspace id.

### DIFF
--- a/lib/models/slack-workspace.js
+++ b/lib/models/slack-workspace.js
@@ -12,7 +12,7 @@ const encrypted = EncryptedField(Sequelize, process.env.STORAGE_SECRET);
 module.exports = (sequelize, DataTypes) => {
   const SlackWorkspace = sequelize.define('SlackWorkspace', {
     slackId: {
-      type: DataTypes.BIGINT,
+      type: DataTypes.STRING,
       allowNull: false,
     },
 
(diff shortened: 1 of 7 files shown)


lib/models/slack-workspace.js
```diff
@@ -12,7 +12,7 @@ const encrypted = EncryptedField(Sequelize, process.env.STORAGE_SECRET);
 module.exports = (sequelize, DataTypes) => {
   const SlackWorkspace = sequelize.define('SlackWorkspace', {
     slackId: {
-      type: DataTypes.BIGINT,
+      type: DataTypes.STRING,
```
wilhelmklopp: WHAT HOW DID THIS EVER WORK?!?

Great find 💯 

Note for other viewers that the schema in the db has the right type but for some reason the model here has it wrong. DDL from Postico:

```
-- Table Definition ----------------------------------------------

CREATE TABLE "SlackWorkspaces" (
    id bigint DEFAULT nextval('"SlackWorkspaces_id_seq"'::regclass) PRIMARY KEY,
    "slackId" character varying(255) NOT NULL,
    "createdAt" timestamp with time zone NOT NULL,
    "updatedAt" timestamp with time zone NOT NULL,
    secrets bytea
);

-- Indices -------------------------------------------------------

CREATE UNIQUE INDEX "SlackWorkspaces_pkey" ON "SlackWorkspaces"(id int8_ops);
CREATE INDEX slack_workspaces_slack_id ON "SlackWorkspaces"("slackId" text_ops);
```

PR #771, opened 2019-01-22
Improve displaying deployments with beta statuses

sburns:
👋 

This improves how deployments in the two new beta statuses (`queued` and `in_progress`) are displayed. More information available about these statuses in the [Deployment Status documentation](https://developer.github.com/v3/repos/deployments/#create-a-deployment-status).

This lumps the two statuses into the same color as `pending` which seems appropriate to me.


wilhelmklopp:
@sburns thank you so much for adding support for the new deployment statuses! Much appreciated 🙇

PR #779, opened 2019-02-06
Debug findOrCreate

gimenete:
Wrap `findOrCreate` to add additional logging information if it fails.


gimenete:
@wilhelmklopp haha yeah. I'm going to create a "cleanup" issue for this and for removing an old secret that we rotated for cookies.

I'm going to write a quick test to fix the coverage check run and I'll deploy this.

PR #790, opened 2019-02-15
Additional debugging for some destroyed subscriptions

gimenete:
Add additional debugging information when destroying a subscription for a permanent error. This way we will be able to investigate issues some users are reporting.

We are getting too many logs of this type:

> Error: An API error occurred: no_permission

and

> Error: An API error occurred: is_archived

@wilhelmklopp I'm wondering if we could have more detailed information from Slack about the error. Any idea?

### DIFF
--- a/lib/activity/router.js
+++ b/lib/activity/router.js
@@ -80,7 +80,7 @@ module.exports = function route(callback) {
           await callback(context, subscription, slack);
         } catch (err) {
           if (isPermanentError(err)) {
-            context.log.info({ err }, 'Permanent error from Slack. Removing subscription');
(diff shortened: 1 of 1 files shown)


lib/activity/router.js
```diff
@@ -80,7 +80,7 @@ module.exports = function route(callback) {
           await callback(context, subscription, slack);
         } catch (err) {
           if (isPermanentError(err)) {
-            context.log.info({ err }, 'Permanent error from Slack. Removing subscription');
+            context.log.info({ err, subscription, payload: context.payload }, 'Permanent error from Slack. Removing subscription');
```
wilhelmklopp: I think `context.payload` might be a bit data heavy here, should we only log the parts out of that we care about like the repo name with owner?

Issue #386, opened 2018-02-27
List enabled features on subscribed repositories


daneah:
It is interesting to know not only which repositories a channel is subscribed to, but also which features of each of those repositories are enabled for notifications. This minimizes people running redundant commands "just to make sure" a feature is enabled.

`/github subscribe list` could potentially do this by default, or `/github subscribe list features` or similar could be added.

cainlevy:
We initially disabled the new GH bot due to a lack of insight into its features. This would help considerably.

Relequestual:
Does the colsure of the associated milestone mean this is no longer priorities work?

kaanchan:
This feature is working :)

megagreg72:
`/github subscribe list features` now exists and works as expected 👍

PR #807, opened 2019-03-04
Add /github subscribe list features

wilhelmklopp:
Another long overdue feature: Listing the settings for each subscription as part of `/github subscribe list features`.

Looks like this:
image

Closes https://github.com/integrations/slack/issues/386


lib/messages/subscription-list.js
```diff
@@ -1,10 +1,19 @@
...
     this.repositories = repositories;
     this.accounts = accounts;
+    this.showFeatureInformation = showFeatureInformation;
+  }
+
+  getFormattedSubscriptionSettings(githubId, type) {
+    return this.subscriptions.find(subscription => (
+      subscription.type === type && subscription.githubId.toString() === githubId.toString()
+    ))
+      .getEnabledSettings().map(setting => `\`${setting}\``).join(', ');
```
wilhelmklopp: using the same logic used here: https://github.com/integrations/slack/blob/42e9489f348adcd6e156bbb81f3c683395b32def/lib/messages/flow/updated-settings.js#L10-L12

lib/commands/list-subscriptions.js
```diff
@@ -26,12 +26,16 @@ module.exports = async (req, res, next) => {
...
   const accounts = await getAccounts(subscriptions.filter(s => s.type === 'account'), robot);
 
-  if (command.text === 'list') {
-    return command.respond(new SubscriptionList(repositories, accounts));
+  if (command.text === '') {
+    // When a user types the following incomplete command /github subcribe
+
+    const response = (new Help(command.command, command.subcommand)).toJSON();
+    const list = new SubscriptionList(subscriptions, repositories, accounts);
+    response.attachments.push(list.toJSON().attachments[0]);
```
gimenete: To make this cleaner I would:

- Make a method in `SubscriptionList` that returns `list.toJSON().attachments[0]`.
- Get the result and pass that as a new constructor argument to `Help`

wilhelmklopp: This is a great suggestion! Thank you 💯

PR #815, opened 2019-03-19
Remove slackEvent case from signed params OAuth handling

wilhelmklopp:
I just discovered that these lines are no longer in use as of https://github.com/integrations/slack/pull/598 so we should be able to get rid of them


wilhelmklopp:
Code coverage was telling us alllll along:
image

https://codecov.io/gh/integrations/slack/src/master/lib/github/oauth.js

PR #816, opened 2019-03-19
Simplify signed params content

wilhelmklopp:
Simplifying and refactoring signed params that are used for completing the GitHub OAuth flow.

Based on the suggestion here: https://github.com/integrations/slack/pull/568#discussion_r186723249

Closes https://github.com/github/ecosystem-primitives/issues/228

### DIFF
--- a/lib/commands/signout.js
+++ b/lib/commands/signout.js
@@ -15,7 +15,10 @@ module.exports = async (req, res) => {
   await signout(robot, slackUser, slackWorkspace);
 
   const state = new SignedParams({
-    trigger_id: command.trigger_id,
+    teamSlackId: command.team_id,
+    userSlackId: command.user_id,
+    channelSlackId: command.channel_id,
+    replaySlashCommand: false,
   });
 
   const { protocol, host } = getProtocolAndHost(req);
--- a/lib/github/oauth.js
+++ b/lib/github/oauth.js
@@ -23,9 +23,17 @@ module.exports = (robot) => {
   });
 
   robot.route().get('/github/oauth/callback', async (req, res) => {
-    // @todo: handle condition where state is expired or has been tampered with
-    // Make sure there is a test for the condition where the state has been tampered with
-    const state = await SignedParams.load(req.query.state);
+    let state;
+    try {
+      state = await SignedParams.load(req.query.state);
+    } catch (err) {
+      if (err.name && err.name === 'JsonWebTokenError') {
+        return res.status(400).send(`Error: ${err.message}`);
+      } else if (err.name && err.name === 'TokenExpiredError') {
+        return res.status(400).send(`Error: ${err.message}`);
+      }
+      throw err;
+    }
 
     const { teamSlackId, userSlackId, channelSlackId } = state;
 
@@ -83,12 +91,16 @@ module.exports = (robot) => {
       attachments: [new SignInConfirm(userSlackId, user.html_url, user.login).getAttachment()],
     });
 
+    const redirectUrl = `https://slack.com/app_redirect?team=${teamSlackId}&channel=${channelSlackId}`;
+
     if (state.replaySlashCommand) {
       const command = await PendingCommand.find(state.trigger_id);
 
       if (!command) {
-        // FIXME: handle issue with expired command
-        throw new Error();
+        // In case it has been more than 30min since the user
+        // executed the initial command, we won't have it in the cache anymore
+        // so we just send them back to Slack without replaying the command
+        return res.redirect(redirectUrl);
       }
 
       return res.redirect(`/slack/command?trigger_id=${command.trigger_id}`);
@@ -98,8 +110,6 @@ module.exports = (robot) => {
       // attempt to fulfil request to unfurl link now that the user is connected
       await deliverAfterSignIn(state.actionCallbackId.replace('unfurl-', ''));
     }
-
-    const redirectUrl = `https://slack.com/app_redirect?team=${teamSlackId}&channel=${channelSlackId}`;
 
     return res.redirect(redirectUrl);
   });
--- a/lib/github/oauth.js
+++ b/lib/github/oauth.js
@@ -24,29 +24,12 @@ module.exports = (robot) => {
 
   robot.route().get('/github/oauth/callback', async (req, res) => {
     // @todo: handle condition where state is expired or has been tampered with
+    // Make sure there is a test for the condition where the state has been tampered with
     const state = await SignedParams.load(req.query.state);
 
-    let command;
-    let teamSlackId;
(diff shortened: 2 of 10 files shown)


lib/github/oauth.js
```diff
@@ -24,29 +24,12 @@ module.exports = (robot) => {
 
   robot.route().get('/github/oauth/callback', async (req, res) => {
     // @todo: handle condition where state is expired or has been tampered with
+    // Make sure there is a test for the condition where the state has been tampered with
```
wilhelmklopp: Still need to do this

lib/github/oauth.js
```diff
@@ -94,31 +77,30 @@ module.exports = (robot) => {
...
       attachments: [new SignInConfirm(userSlackId, user.html_url, user.login).getAttachment()],
     });
-    if (state.slackAction && /unfurl-\d+/.test(state.slackAction.callback_id)) {
+
+    if (state.replaySlashCommand) {
+      const command = await PendingCommand.find(state.trigger_id);
+
+      if (!command) {
+        // FIXME: handle issue with expired command
+        throw new Error();
```
scepticulous: Not sure if the FIXME will be addressed in this PR. If not, I wonder if we could pass a little bit more information as part of the exception.

lib/commands/signout.js
```diff
@@ -15,7 +15,10 @@ module.exports = async (req, res) => {
   await signout(robot, slackUser, slackWorkspace);
 
   const state = new SignedParams({
-    trigger_id: command.trigger_id,
```
scepticulous: Just a general question: Do we have conventions around omitting the key vs setting it's value to `null` like: `trigger_id: null` ? If not, I would like to have your, and/or the teams perspective on it. Just to get to know how we work.

lib/middleware/require-github-user.js
```diff
@@ -11,7 +11,12 @@ module.exports = async function requireGitHubUser(req, res, next) {
   const { slackUser } = res.locals;
 
   if (!slackUser.GitHubUser) {
-    const state = new SignedParams({ slackAction: { ...req.body, token: undefined } });
```
gimenete: What was the reason to have `token: undefined` and why it is no longer needed?

gimenete:
Great work 🎉 

I just left a comment. Everything else looks good. I know that it still needs a bit of work (e.g. additional test(s)) so I'll do an additional review when everything is finished.

lib/middleware/require-github-user.js
wilhelmklopp: This was solely a measure to not leak the `token` (which is part of `req.body`) 🙂

lib/github/oauth.js
wilhelmklopp: Good shout! Yeah I ended up just addressing this too in this PR 😄

lib/commands/signout.js
wilhelmklopp: It's a really good question, I don't think we have a very clear pattern/rule for this across the codebase. Do you have any thoughts as to what a good pattern might look like?

As extra context, `trigger_id` should only be needed now when we also pass `replaySlashCommand`. If we were using Typescript, there would probably be a nice way to express this in the code 😄

lib/github/oauth.js
```diff
@@ -23,30 +23,21 @@ module.exports = (robot) => {
...
+        return res.status(400).send(`Error: ${err.message}`);
       }
-      teamSlackId = command.team_id;
-      channelSlackId = command.channel_id;
-      userSlackId = command.user_id;
-    } else if (state.slackAction) {
-      teamSlackId = state.slackAction.team.id;
-      userSlackId = state.slackAction.user.id;
-      channelSlackId = state.slackAction.channel.id;
+      throw err;
```
wilhelmklopp: This is the only line that's not covered now and I'm not sure if it should or even can be covered in a clean way? Having to manually ensure that we're only `catch`-ing the errors we care about is a bit of JS-language flaw.

wilhelmklopp:
@gimenete (and @scepticulous if you feel so inclined) this should be ready for re-review now 😸

lib/github/oauth.js
gimenete: What if you do all the error handling for known errors in the `SignedParams.load()` function and you return an object like `{ err, state }` and then here just do

```
try {
  const { err, state } = await SignedParams.load(req.query.state);
  if (err) res.status(400).send(`Error: ${err.message}`); // handled error
} catch (err) {
  throw err; // unhandled error
}
```

Then you can unit test `SignedParams.load` for `JsonWebTokenError` and `TokenExpiredError` errors.

wilhelmklopp: In this case, wouldn't the `  throw err; // unhandled error` still be uncovered? 😄 

For what it's worth the `JsonWebTokenError` and `TokenExpiredError` errors _are_ already covered now: https://github.com/integrations/slack/pull/816/files#diff-07e1c12a9232d0340549c220b7392130R130-R183

lib/middleware/require-github-user.js
scepticulous: I like making sure, it is definitely not leaked from a security perspective. However I would prefer using `null` whenever an assignment to "nothing" is intended, so that nobody confuses that with the value not being assigned at all. But that's just opinion and at most something to consider in the future.

wilhelmklopp:
> But I'm going to do another suggestion about the error test. Maybe you can export the whole async (req, res) => {.. function and test it using a mock req and res. If the req is an empty object you'll get an error thrown in req.query.state becuase query is undefined. It wouldn't even invoke SignedParams.load() so the test should be pretty simple.

@gimenete could you expand on this a bit more? 😄  We should already be able test this end to end using the simulated request pattern we have in integration tests (example: https://github.com/integrations/slack/pull/816/files#diff-07e1c12a9232d0340549c220b7392130R177-R179), but it's quite possible that I'm missing something 😅

gimenete:
@wilhelmklopp I was referring to your comment:

> This is the only line that's not covered now and I'm not sure if it should or even can be covered in a clean way?

https://github.com/integrations/slack/pull/816#discussion_r268126212

I see that you are covering the known error cases and there's a missing case: when there's an unknown error and thus, `throw err` is executed.

What I'm suggesting is doing something like:

```js
const handler = async (req, res) => {
  ...
}

obot.route().get('/github/oauth/callback', handler)

// export handler so it can be unit tested

// Then in the test:
test("error handler". () => {
  const req = {} // it doesn't have a query attribute
  expect(() => handler(req)).toThrow() // complete this with assertions for error message
})

wilhelmklopp:
@gimenete ahhh, sorry, got it! I'm going to open a follow up PR for this 👍

PR #818, opened 2019-03-21
Improve comment cache key

wilhelmklopp:
Based on recommendation outlined in https://github.com/github/ecosystem-primitives/pull/229

### DIFF
--- a/lib/activity/comments.js
+++ b/lib/activity/comments.js
@@ -31,7 +31,7 @@ module.exports = async (context, subscription, slack) => {
     }).getRenderedMessage(),
   ];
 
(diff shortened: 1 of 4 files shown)


scepticulous:
Not really sure, but do you also have to update the [snapshot tests](https://github.com/integrations/slack/blob/master/test/integration/__snapshots__/notifications.test.js.snap#L22-L37) ?

wilhelmklopp:
@scepticulous yep I think those snapshots were updated here: https://github.com/integrations/slack/pull/818/files#diff-f47210308778f23b45cd1f8581643d05

PR #820, opened 2019-03-21
Scope channel model caching by workspace

wilhelmklopp:
Based on recommendation outlined in github/ecosystem-primitives#229

### DIFF
--- a/lib/slack/channel.js
+++ b/lib/slack/channel.js
@@ -1,15 +1,16 @@
 module.exports = class Channel {
   constructor({
-    id, cache, client, logger,
+    id, workspaceId, cache, client, logger,
   }) {
     this.id = id;
+    this.workspaceId = workspaceId;
     this.cache = cache;
     this.client = client;
     this.logger = logger;
   }
 
   cacheKey(...parts) {
-    return [`channel#${this.id}`].concat(parts).join(':');
+    return [`channel#${this.workspaceId}#${this.id}`].concat(parts).join(':');
   }
 
   async post(message) {
--- a/test/integration/index.js
+++ b/test/integration/index.js
@@ -38,7 +38,7 @@ beforeEach(() => {
 
   // Clear all data out of the test database
   return Promise.all([
-    models.sequelize.truncate({ cascade: true }),
+    models.sequelize.truncate({ cascade: true, restartIdentity: true }),
     cache.clear(),
   ]);
 });
--- a/lib/commands/issue-state.js
+++ b/lib/commands/issue-state.js
@@ -8,6 +8,7 @@ module.exports = state => async (req, res) => {
 
   const channel = new Channel({
     id: command.channel_id,
+    workspaceId: slackWorkspace.id,
     cache,
     client: slackWorkspace.client,
     logger: req.log,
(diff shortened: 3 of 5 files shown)


test/integration/index.js
```diff
@@ -38,7 +38,7 @@ beforeEach(() => {
 
   // Clear all data out of the test database
   return Promise.all([
-    models.sequelize.truncate({ cascade: true }),
+    models.sequelize.truncate({ cascade: true, restartIdentity: true }),
```
wilhelmklopp: I'm not sure how our tests ever worked well without this (especially the cache key snapshots), but either way this should help make our test suite a bit more deterministic.

cc @gimenete

lib/slack/channel.js
```diff
@@ -1,15 +1,16 @@
...
     this.id = id;
+    this.workspaceId = workspaceId;
     this.cache = cache;
     this.client = client;
     this.logger = logger;
   }
 
   cacheKey(...parts) {
-    return [`channel#${this.id}`].concat(parts).join(':');
+    return [`channel#${this.workspaceId}#${this.id}`].concat(parts).join(':');
```
scepticulous: nice. I think the scoping is a great idea. I don't know if it's better, more accurate or consistent to do either:
`ws:#{this.workspaceID}#channel#[...]` or `channel#<wsid>#<channelid>`. Just want to make sure we are doing it this way because it makes sense and not to keep the identifier short.

test/integration/index.js
gimenete: Hmmm, cache keys have always been based on Slack and GitHub ids as far as I understand. Those ids are written by hand in tests and fixtures. That's why they've worked.

I see that you are now using the `workspaceId`. Can we get an id from Slack for this so we make our tests not dependant on ids based on sequences?

wilhelmklopp: > Hmmm, cache keys have always been based on Slack and GitHub ids as far as I understand.

From what I can tell most cache keys that cache the workspace id use our database id. I included cache key examples in my report here: https://github.com/github/ecosystem-primitives/pull/229/files?short_path=0d9829a#diff-0d9829a4a83bd19741db3959bf362d99

I believe the logic for all of those is this method:
https://github.com/integrations/slack/blob/cb787ec09865b518410fc1f4e9d2e34466018667/lib/models/subscription.js#L104-L106

lib/slack/channel.js
wilhelmklopp: Yeah, I agree it would be even better if we had a `ws:` or something like that as part of this key.

In terms of consistency, I tried to make it similar to the other keys that cache by workspace and channel, examples here: https://github.com/github/ecosystem-primitives/pull/229/files?short_path=0d9829a#diff-0d9829a4a83bd19741db3959bf362d99

And I didn't want to update the structure of all of them to include `ws` 😬

test/integration/index.js
gimenete: Hmmm ok.

At some point in the future I think we should make `SlackWorkspaces.id` the real Slack id and here `slackWorkspaceId` would be also the Slack id, and not an internal id.

wilhelmklopp: yeah i think that makes sense, though I don't look forward to that migration haha 👍

gimenete: hahaha, yeah, it's a tough one 😜

PR #831, opened 2019-04-01
Show interstitial page before connecting GitHub account

wilhelmklopp:
As discussed, we would like to show an interstitial page before redirecting the user to go through the GitHub OAuth flow.

Todo: 
- [x] Improve user facing error message when JWT is expired
- [x] get diff test coverage to 100%

Closes https://github.com/github/ecosystem-primitives/issues/249


wilhelmklopp:
This is what it looks like 🙌 

image

Grateful for any copy improvements suggestions  😄 

cc @tcbyrd

views/confirm-login.hbs
```diff
@@ -0,0 +1,48 @@
...
+    gtag('config', 'UA-3769691-98');
+  </script>
+</head>
+
+<body>
+
+  <div class="text-center">
+    <div class="container-lg p-responsive py-6 py-lg-10">
+      <img src="/static/img/githubplusslack.svg" class="mb-3 mb-sm-5 col-6 col-sm-12" style="max-width: 300px;">
+      <p>Finish connecting your GitHub account to your slack account on </p>
```
tcbyrd: Should Slack be capitalized here? Not sure what their branding guidlines are, but since we capitalized GitHub I thought it should be.

```suggestion
      <p>Finish connecting your GitHub account to your Slack account on </p>
```

views/confirm-login.hbs
```diff
@@ -0,0 +1,48 @@
...
+      <p class="alt-lead text-gray col-md-8 mx-auto mb-3 mb-lg-4">
+        {{teamSlackDomain}}.slack.com
+      </p>
+      <p>You will be able to view rich previews for GitHub links, open and close issues, and much more from Slack</p>
```
tcbyrd: I'm thinking we might want to be explicit here on why this is required beyond just installing the app. It's not required to do this to view public GitHub links, right? So 👇 suggesting adding `private` and adding a bit about how it depends on what access you have on GitHub:

```suggestion
      <p>This will connect your accounts so you can use Slack to view rich previews for private GitHub links, open and close issues, and other features that depend on your access to GitHub!</p>
```

lib/frontend.js
```diff
@@ -12,4 +16,55 @@ app.get('/', (req, res) => {
...
+    teamSlackId,
+    teamSlackDomain,
+    userSlackId,
+    channelSlackId,
+    replaySlashCommand,
+    actionCallbackId,
+  } = state;
+
+  // Set session cookie with oauth state that we can verify it in the OAuth callback
+  const githubOAuthState = crypto.randomBytes(Math.ceil(30 / 2)).toString('hex').slice(0, 30);
```
gimenete:
```suggestion
  const githubOAuthState = crypto.randomBytes(15).toString('hex');
```

Since `Math.ceil(30 / 2)` is constant, why not putting it directly? 😄  Also I've checked that the result is always a 30 chars string, so there's no need to do the `slice()`. Also, interesting use of `slice()`, I would have used `substring()` since it's a string 🤔

scepticulous: Not being the author, I guess the intend was documentation, because one can see how the 30 and the 30 match. So it reads like: "we want 30 chars, first we take 15 random bytes, then we hex encode them to make them 30 (since each byte is 00..FF) and in order to guarantee that we always create a valid token, we slice 30, no matter what" ... It's a nice story that the code is telling. However, it doesn't really make sense, so 
`crypto.randomBytes(15).toString('hex');` is way better 😄

scepticulous: > Also, interesting use of `slice()`, I would have used `substring()` since it's a string 🤔
Fun fact, I think `slice` has the more appropriate behaviour. `substring` is marginally faster than slice (depending on browser details). 
Here is a comparison: https://stackoverflow.com/a/2243835/2375386. I think the argument swapping of substring is horrible. But its just opinion, so never mind. 🤷‍♂️

views/confirm-login.hbs
wilhelmklopp: So much better, thanks @tcbyrd ✨

lib/frontend.js
wilhelmklopp: All great points 💯 

For context, this originally came from here: https://github.com/integrations/slack/pull/264#discussion_r167062267

wilhelmklopp:
> Does that sound accurate?

Yep 💯 

> If so, I think we're good to go here but I did want to point out that we may not have sufficient protection against homoglyph attacks so it may be possible to create a slack team domain that looks close to your victim's but actually isn't (I've not performed live testing to validate).

I think Slack is _somewhat_ restrictive here with respect to what characters are allowed in a team domain, but I agree that this would be good to validate 👍 


> Also, any XSS on this page (which there is nothing obvious that indicates it is vulnerable) could result in someone being automatically redirected so it may make sense to open a separate issue to implement a Content Security Policy. Should I open up that issue?

Great point! Yes, if you could open an issue for this that would be awesome 🤘

PR #839, opened 2019-04-12
Fixes signature verification for ansi codes

scepticulous:
Updates the requires version of probot, which includes an update of octokit/webhook.js that fixes signature verification for payloads that contain ANSI codes. In particular `ESC` sequences.

See:
* https://github.com/octokit/webhooks.js/issues/71
* https://github.com/octokit/webhooks.js/issues/73
* https://github.com/probot/probot/pull/912

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -334,11 +334,10 @@
       }
     },
     "@octokit/webhooks": {
-      "version": "5.3.2",
-      "resolved": "https://registry.npmjs.org/@octokit/webhooks/-/webhooks-5.3.2.tgz",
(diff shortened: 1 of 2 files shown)


gimenete:
As far as I understand `package-lock.json` should contain new information too.

scepticulous:
> As far as I understand `package-lock.json` should contain new information too.

🤦‍♂️ Yes of course. I really need to get used to the fact, that we are not using yarn. I use it locally because it's so much faster and nicer and I noticed that yarn.lock was not checked in and completely forgot that we are using npm... I updated the `package-lock.json` now. Thanks for reminding me 👍

PR #857, opened 2019-06-04
Update uses of deprecated methods

tarebyte:
This is a first step to upgrade the Slack app to the latest version of Probot.

The PR removes the method deprecation warnings I found while running the test suite, I'm going to look to see if there are more that I missed from the changelogs.


IAmHughes:
This looks like a great start @tarebyte, thank you!

ahmad380360:
شكرا

Issue #862, opened 2019-06-07
How do I get subscribe to listen to "checks" in a PR?


mastoj:
We are moving more and more things over to github action, and one thing we want to have then is a notification in slack whenever a build finishes, either success or fail. 

I thought subscribing to the `statuses` would be enough, but nothing is published to the slack channel when a check finishes or fails.

Have I misunderstood the use of `/github subscribe owner/respository statuses`?

mastoj:
Reading here I start to believe that maybe I need to post to the statuses api myself: https://developer.github.com/v3/repos/statuses/, is that correct?

mastoj:
So `statuses` and `checks` are different I realized, but I still can't get it working. I do manage to post `statuses` updates to github using the api (which I verify worked through the api).

I have tried to subscribe to statuses by running the following:

    /github subscribe list <owner>/<repo> statuses 

But nothing is updated when I do `POST` a status update.

IAmHughes:
Hey @mastoj , just wanted to let you know that the integration now supports displaying checks in the PR messaging you see in Slack. You can read about the changes we shipped today [on the GitHub Blog](https://github.blog/2019-07-29-whats-new-in-the-github-and-slack-integration/). 

The key thing to note is that this required some updated permissions on the application, so the owner of any repositories that have the GitHub Slack App installed will need to accept those permissions. If you try to use the command without the permissions, you'll be prompted to update them ❤️.

Thanks for your patience on this!

villesau:
@IAmHughes is the feature deprecated? I'm not able to subscribe to check statuses in Slack.

<img width="731" alt="Screenshot 2022-08-09 at 15 28 34" src="https://user-images.githubusercontent.com/2647016/183646898-d1f016c9-65f1-4402-857a-f9a844ed968f.png">

david-wolgemuth:
> Uh oh! `statuses` is not a feature. Available features are:
> `issues`, `pulls`, `commits`, `releases`, `deployments`, `reviews`, `comments`, `branches`, `+label:"my label"`, `discussions`, `workflows`

Issue #751, opened 2018-12-20
Subscribing to events from the Checks API


dentarg:
Doesn't [look like](https://github.com/integrations/slack/blob/04a2cb6de58b705de5c379fff34000c4d26c1270/README.md#repository-activity) it is possible to subscribe to events from the Checks API.

It would be great to be able to do.

This documentation, https://developer.github.com/v3/activity/events/types/, might be relevant.

IAmHughes:
Hey @dentarg , just wanted to let you know that the integration now supports displaying checks in the PR messaging you see in Slack. You can read about the changes we shipped today [on the GitHub Blog](https://github.blog/2019-07-29-whats-new-in-the-github-and-slack-integration/). It's not exactly subscribing just to checks, but it may help unblock your use case.

The key thing to note is that this required some updated permissions on the application, so the owner of any repositories that have the GitHub Slack App installed will need to accept those permissions. If you try to use the command without the permissions, you'll be prompted to update them ❤️.

Thanks for your patience on this! Please re-open this issue if you find that there is an additional use case for subscribing to checks specifically in another way. If there is, if you can give an example of what you'd like to see that would be great!

dentarg:
@IAmHughes I think the use case I had in mind when opening this issue was to get feedback from Travis CI [Cron Jobs](https://docs.travis-ci.com/user/cron-jobs/) in Slack (without having to set up Travis -> Slack integration). However, I'm not sure it is possible to use the Checks API without the context of a pull request. Do you know? :)

IAmHughes:
@dentarg ah okay that's an interesting use case and a bit different than what we shipped. I think this latest update will help some, in that the Checks Suites (see the `checks` tab of a pull request) will report their status when a pull request is unfurled, opened, etc.

- `Checks` tab:
![image](https://user-images.githubusercontent.com/2894107/62218722-9e00b180-b372-11e9-997f-52eef3706ff1.png)


- New functionality:
![image](https://user-images.githubusercontent.com/2894107/62218636-7873a800-b372-11e9-94b2-7d9d8d6021f9.png)

I'll re-open this issue as what you're referring to would be a little different I believe. Thanks for the feedback!

scepticulous:
Hi @dentarg 👋 

>  However, I'm not sure it is possible to use the Checks API without the context of a pull request. Do you know? :)

[...]

PR #863, opened 2019-06-11
Feature/checks api

imjohnbo:
The GitHub + Slack app currently reports Status updates on open pull requests, and now we're expanding support to Checks as well.  🎉

After subscribing to a GitHub repository (`/github subscribe owner-or-repository`), you will see status updates on pull requests opened in the last 30 minutes from integrations that use the [Checks API](https://developer.github.com/v3/checks/) _and_ those that are still using [Statuses](https://developer.github.com/v3/repos/statuses/):

image

image


lib/messages/status.js
```diff
@@ -4,9 +4,13 @@ const {
 } = require('.');
 
 class Status extends Message {
-  constructor(status) {
+  constructor(checkOrStatus) {
     super({});
-    this.status = status;
+    this.status = checkOrStatus.state || checkOrStatus.conclusion || checkOrStatus.status;
```
gimenete: The `checkOrStatus.conclusion` thing is confusing here but I get what you're doing because for check runs the most important thing is the `conclusion` (e.g. `success`). It was confusing because you're storing this in a `this.status` variable. Maybe we could use a better name for that variable, but in any case looks good to me 👍

lib/messages/pull-request.js
```diff
@@ -106,32 +124,24 @@ class PullRequest extends AbstractIssue {
       };
     }
 
-    const attachments = [
+    let attachments = [
```
gimenete: In JavaScript `const` doesn't mean _immutable_, it just means that the variable cannot be reassigned. So you can keep using `const` here.

IAmHughes:
This should be related to #751 and #862, potentially being able to close those ones out 🎉 ✨ 💖!!

Mentioning them for cross-linking ❤️

lib/messages/status.js
imjohnbo: Agreed about it being confusing! How's `combinedStatus` or `state`?

lib/messages/status.js
```diff
@@ -20,28 +25,37 @@ class Status extends Message {
...
+      color,
       fallback,
       text: `:white_check_mark: ${fallback}`,
       footer: `${successfulChecks}/${totalChecks} successful checks`,
     };
   }
 
   renderAttachment() {
-    const summary = `${this.status.context}: ${this.status.description}`;
+    const summary = `${this.context}: ${this.description}`;
```
IAmHughes: I love this new way we're building Statuses for attachments on messages!

Just for due-diligence, I believe #866 could be fixed in this new format by adding a `null` check to the `description` field here. I didn't do deep testing yet, but I think it could be a quick `null` check to fix it!

imjohnbo: We're thinking the same thing, @IAmHughes! 👍

wilhelmklopp:
So cool to see all of this work! 😎 Would you mind adding a couple of screenshots in the PR description?

And (in case there were any) tradeoffs or any considerations of note? 🙇

lib/messages/pull-request.js
```diff
@@ -109,29 +126,25 @@ class PullRequest extends AbstractIssue {
     const attachments = [
       this.getMainAttachment(),
     ];
-    if (this.statuses) {
-      // 3 cases: none or one successful, some successful, all successful
```
wilhelmklopp: Do you reckon it would be useful to keep one summary comment at the top of this logic like the one we had before?

lib/messages/status.js
```diff
@@ -4,15 +4,25 @@ const {
...
 class Status extends Message {
-  constructor(status) {
+  constructor(checkOrStatus) {
     super({});
-    this.status = status;
+    /*  error, failure, pending, success, queued, in_progress,
+        completed, neutral, cancelled, timed_out, or action_required
+    */
+    this.overallStatus = checkOrStatus.state || checkOrStatus.conclusion || checkOrStatus.status;
+    this.context = checkOrStatus.context || checkOrStatus.app.name || checkOrStatus.name;
```
wilhelmklopp: Would we ever get to that third conditional here?

We should always either have a `checkOrStatus.context` or `checkOrStatus.app.name`, right? 😅

wilhelmklopp: Actually, on second thought isn't `context` also always required? haha

lib/models/subscription.js
```diff
@@ -5,6 +5,7 @@ module.exports = (sequelize, DataTypes) => {
   const EventNameToSettings = {
     pull_request: 'pulls',
     status: 'statuses',
+    check_run: 'statuses',
```
wilhelmklopp: 👍

lib/activity/pull-requests.js
```diff
@@ -59,53 +59,89 @@ function storePRMapping(context) {
...
+}
+
+async function getChecksListForRef(context, sha) {
+  return (await context.github.checks.listForRef(context.repo({
+    ref: sha,
+  }))).data.check_runs;
+}
+
+async function getPullRequest(context, message) {
+  return (await context.github.pullRequests.get({ ...message.context })).data;
```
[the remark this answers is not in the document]
imjohnbo: `message.context` has the **GitHub** owner, repo, and number, but I'm happy to pull it from elsewhere if you think it's better!

lib/messages/status.js
imjohnbo: yeah that's probably true, I was going overboard to make sure we fixed #866

lib/activity/pull-requests.js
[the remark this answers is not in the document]
wilhelmklopp: Yeah I think that would be my preference, so that for future readers of this code `message` continues to refer to the cached Slack **message**, as opposed to other data that is also cached with it.

Mostly a variable naming/clarity nitpick 😅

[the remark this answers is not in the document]
imjohnbo: Hmm I do like clarity, but aren't we specifically caching these values so we can easily retrieve a pull request like this later? I'm having trouble finding where to get `pull_number` if not from the cache, and `{ ...message.context }` becomes a little uglier with something like

```
{
    owner: context.payload.repository.owner.login,
    repo: context.payload.repository.name,
    pull_number: ?
}
```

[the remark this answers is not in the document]
wilhelmklopp: My bad! I think i've got something backwards. Looks like this is not actually a new change in this PR and we were always storing PR metadata alongside the Slack mesage?

I think I was imagining that we could get the pull request directly by ID, which is cached separately: https://github.com/integrations/slack/blob/520dac1b47611c1276e26fedf37a2e466d42df7b/lib/activity/pull-requests.js#L64

But that improvement can definitely happen outside of this PR! Sorry 😓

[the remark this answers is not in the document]
imjohnbo: No worries! I'll keep it focused and unchanged for now by using `message.context`.

lib/activity/pull-requests.js
```diff
@@ -59,53 +59,90 @@ function storePRMapping(context) {
...
+  });
+  context.log(res, 'Updated Slack message');
+}
+
+async function onStatusOrCheckRun(context, subscription, slack) {
+  let sha;
+  let id;
+
+  if (context.event === 'status') {
+    ({ sha } = context.payload);
```
gimenete: I didn't know this was even possible 😛 !!

lib/messages/pull-request.js
```diff
@@ -23,7 +24,27 @@ class PullRequest extends AbstractIssue {
...
+      this.successfulStatuses.push(...statuses.filter(status => status.state === 'success'));
+    }
+
+    if (checkRuns) {
+      this.unsuccessfulStatuses.push(...checkRuns.filter(checkRun =>
+        checkRun.conclusion !== 'success' &&
+        checkRun.conclusion !== 'neutral'));
+      this.successfulStatuses.push(...checkRuns.filter(checkRun =>
+        checkRun.conclusion === 'success' ||
+        checkRun.conclusion === 'neutral'));
```
gimenete: We could improve this a little bit if we had the conclusions in an array maybe called `successfulConclusions` and then do `successfulConclusions.includes(checkRun.conclusion)` but it's ok like it is now too 👍

lib/messages/status.js
```diff
@@ -4,15 +4,24 @@ const {
...
+      checkOrStatus.name;
+    this.avatar_url = checkOrStatus.avatar_url || checkOrStatus.app.owner.avatar_url;
+    this.target_url = checkOrStatus.target_url || checkOrStatus.html_url;
   }
 
   static getStatusColor(status) {
     if (status === 'success') {
       return constants.STATUS_SUCCESS;
-    } else if (status === 'pending' || status === 'in_progress' || status === 'queued') {
+    } else if (status === 'pending' || status === 'in_progress' || status === 'queued' || status === 'neutral') {
```
gimenete: Same than the previous comment. It could be a bit better to have these statuses in an array and then use `.includes()` but again, it's ok to leave it like it is by now.

PR #865, opened 2019-06-13
Add `app.yml` and permissions/events to `CONTRIBUTING.md`

IAmHughes:
While setting up my local dev environment, I noticed the `CONTRIBUTING.md` file did not list which permissions and events were needed. This PR addresses this.

Additionally, for first-time contributors, following the [app manifest flow](https://developer.github.com/apps/building-github-apps/creating-github-apps-from-a-manifest/) would be helpful. Therefore, this PR also introduces an `app.yml` file with the permissions and events needed.


IAmHughes:
> I'm curious if when a new contributor starts the app for the first, they will actually see the probot welcome page and the "register GitHub App" button or if there is something in our setup that will prevent that?

That is a good question...I'll have to do some digging in the future 😄

PR #877, opened 2019-06-25
Upgrade Probot dependency to the latest version

tarebyte:
This update Probot to the latest version 9.2.15, so that we can take advantage of Redis caching.


IAmHughes:
One thing I'd like to do once this is ready to be merged is to look for any open issues/PRs from Greenkeeper and such that call out version updates. Since a lot of dependencies may have been updated/impacted, there may be some cleanup we can do.

lib/api.js
```diff
@@ -17,7 +17,12 @@ module.exports = (robot) => {
     }
 
     const github = GitHubAPI({
+      Octokit: ProbotOctokit,
       logger: req.log,
+      retry: {
+        // disable retries to test error states
+        enabled: false,
```
wilhelmklopp: We think that some assumptions in the app will break if we turn on automatic retries, so leaving them disabled for now :)

lib/github/oauth.js
```diff
@@ -63,7 +63,14 @@ module.exports = (robot) => {
...
     const accessToken = queryString.parse(accessTokenResponse.data).access_token;
 
-    const github = GitHubAPI({ logger: req.log });
+    const github = GitHubAPI({
+      Octokit: ProbotOctokit,
+      logger: req.log,
+      retry: {
+        // disable retries to test error states
+        enabled: false,
+      },
```
wilhelmklopp: I think there are a few more places where we need to disable throttling like here ☝️

scepticulous: Thanks for pointing that out. Fixed in https://github.com/integrations/slack/pull/877/commits/809063711c94dbb1b4fe90ba994e6fd190167416

PR #889, opened 2019-07-29
Add ngrok to Brewfile

jnraine:
In order to run the the Slack integration locally, you need to use ngrok. This ensures ngrok is installed by `script/bootstrap`. (This is a small tweak, with more improvements to setup on the way later.)


jnraine:
@scepticulous that's a great point. I'll update this to make ngrok optional.

Is the reason we use ngrok over smee.io?

scepticulous:
For posterity from our slack discussion:
----
Q:
> Is the reason we use ngrok over smee.io?

A:
> @tcbyrd wrote:
> I use smee for probot apps like stale. Jira and Slack require a web front end and smee is just a webhook proxy

Q: 
> What about [localtunnel](https://localtunnel.github.io/www/)

A:
> We built smee.io because localtunnel was unreliable [...], plus probot apps at the time didn’t need a full tunnel
Was intended to be a lightweight way to receive webhooks locally is all
I wouldn’t be against re-evaluating localtunnel. This was 2 years ago at this point

------
There’s also a new service from Cloudflare: https://developers.cloudflare.com/argo-tunnel/downloads/

tcbyrd:
For historical context, here's the discussion on why we ended up building smee: https://github.com/probot/friction/issues/5

ngrok has always seemed to be the most reliable full tunnel solution and generally "just works" when you need something more than just a webhook proxy.

PR #890, opened 2019-07-29
Simplify app startup

jnraine:
In order to use run this integration locally, you need something like ngrok to make it accessible via the internet so webhooks from slack.com and github.com can be received. Prior to this change, developers needed to run ngrok and the server in different tabs.

This has downsides. When getting started, you need to learn a few things to get the app running. If you forget to start ngrok, things won’t work. If you forget to turn off ngrok after you’re done developing, a port on your computer is publicly accessible.

Now just run: `overmind start`

image

This starts the `slack` ngrok tunnel and streams the logs along side the probot server. In order to use this, you need to store an ngrok tunnel. There’s a template you can use included in this commit:

1. `cp ngrok.example.yml ngrok.yml`
2. Update `ngrok.yml` with your token and your handle
3. `cp ngrok.yml ~/.ngrok2/ngrok.yml`

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -14,7 +14,7 @@ const probot = createProbot({
   id: process.env.APP_ID,
   secret: process.env.WEBHOOK_SECRET,
   cert: findPrivateKey(),
-  port: process.env.PORT || 3000,
+  port: process.env.UGH_PORT || process.env.PORT || 3000,
   webhookPath: '/github/events',
   webhookProxy: process.env.WEBHOOK_PROXY_URL,
 });
--- a/Brewfile
+++ b/Brewfile
@@ -1,3 +1,4 @@
+brew "overmind"
 brew "node"
 brew "redis"
 brew "postgres"
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+ngrok: ngrok start -log=stdout slack
+slack: npm run dev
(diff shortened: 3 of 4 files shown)


lib/run.js
```diff
@@ -14,7 +14,7 @@ const probot = createProbot({
   id: process.env.APP_ID,
   secret: process.env.WEBHOOK_SECRET,
   cert: findPrivateKey(),
-  port: process.env.PORT || 3000,
+  port: process.env.UGH_PORT || process.env.PORT || 3000,
```
jnraine: Overmind overrides `PORT` with [some other value](https://github.com/DarthSim/overmind#specifying-the-ports). It's a bit annoying in this scenario. I've added this as a temporary workaround.

Very open to suggestions!

jules2689: What is UGH_PORT?

scepticulous:
I really 💟  the idea of reducing friction. 👍 

Given, that this is an open-source repository and not everybody uses/can use ngrok (see [this comment](https://github.com/integrations/slack/pull/889#pullrequestreview-268232641)), there are a few questions which are related to your initial findings:

> This has downsides. When getting started, you need to learn a few things to get the app running. 

Can we improve the docs, so that we reduce friction for contributors that use `yarn` and/or smee.io (for example)? This way, people that cannot use this setup, can still have an improved experience. 
Another small improvement is using a far less common port per default. I think 3000 is just too common and this way we can make the first experience better. (No need to stop other services and it works without setting PORT)

Despite these related topics, I am really looking forward to this shipping this improvement 👍

PR #891, opened 2019-07-30
Use sorted arrays in cache snapshots

gimenete:
This is a solution or a mitigation to those flaky tests that were depending in the order of execution.

By extracting the content of the Sets and Maps that we use and sorting it, I expect to have a lot less problems with the order of execution.


scepticulous:
After fighting with travis throughout the day to make the build green I noticed one thing in this diff:

https://github.com/integrations/slack/pull/891/files#diff-828aa4eb0b3318112235eff66b07a940L55-R77

```javascript
Map {
  "pending-command#13345224609.738474920.8088930838d88f008e0" => Object {
    "channel_id": "C2147483705",
    [...]
    "user_name": "Steve",
  },
}
```
changed to
```javascript
Array [
  Array [
    "pending-command#13345224609.738474920.8088930838d88f008e0",
    Object {
      "channel_id": "C2147483705",
       [...]
      "user_name": "Steve",
    },
  ],
```
We are Changing a Map of `pending_command => object` to `['pending_command', object]` which does not seem to be affected by sorting since the keys within the object are still sorted the same way. But maybe that is only this snapshot and it does not matter.
However, this led me to believe, I might not really understand the deeper details of the change, which is the reason why I did not ✅ it yet.

gimenete:
@scepticulous This PR is changing `Set`s to sorted arrays and `Map`s to arrays of entries sorted by the key. Each entry is an array on itself with the key as first element and the value as second element. So an array of entries is an array of arrays.

PR #895, opened 2019-08-05
Add global.gc() afterAll and increase max-old-space-size

scepticulous:
While running tests locally I repeatedly ran into memory allocation errors:
`FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`.

I enabled `logHeapUsage` and noticed, that even test failures related to `timeouts` and missing mock requests were GC related.
There are open issues regarding jest [like this](https://github.com/facebook/jest/issues/7311), and when running jest in `watch` mode, that total memory keeps growing. On my system, it consumed more than 4GB after 4-5 full test runs.

In order to address these issues, I added the following changes to the test command invocation:
* `expose-gc` to allow us manually running `global.gc()` afterAll
* `logHeapUsage` for manual inspection
* `max-old-space-size=4096` to increase the old-gen size to 4 Gig instead of 1,5 Gig. This should make testmore reliable with regards to 'out of memory' and 'timeout' errors.
* setting node flags implied adding the path to `jest`, which is good to make sure, that everybody run tests with the same jest version and does not accidentally use a globally installed version.

If your laptop does not support 4GB for the GC, just add `NODE_OPTIONS=--max-old-space-size=1024` before `npm test` and the value will be passed to v8. 
See [allowed v8 flags](https://nodejs.org/dist/latest-v8.x/docs/api/cli.html?source=post_page---------------------------#cli_node_options_options)

This should make tests more reliable and limit the memory allocation in watch mode.

Finally, I extracted the custom `jestTimeout` into a constant instead of having a magic number in many tests.


package.json
```diff
@@ -3,8 +3,8 @@
   "license": "MIT",
   "scripts": {
     "start": "node ./lib/run.js",
-    "test": "jest --coverage --runInBand --forceExit",
-    "test:watch": "jest --runInBand --forceExit --watch --notify",
+    "test": "node --expose-gc --max-old-space-size=4096 ./node_modules/.bin/jest --coverage --runInBand --logHeapUsage --forceExit",
+    "test:watch": "node --expose-gc --max-old-space-size=4096 ./node_modules/.bin/jest --runInBand --logHeapUsage --forceExit --watch --notify",
```
wilhelmklopp: I _believe_ `node_modules/.bin` are automatically part of the `$PATH` thanks to some node/package.json magic, so we might be able to drop that? Buuut I have never done anything this low level with node, so it's very possible that I'm very wrong here 😅

test/integration/deployment.test.js
```diff
@@ -4,11 +4,9 @@ const nock = require('nock');
...
 const fixtures = require('../fixtures');
 
-const {
-  SlackUser,
-  GitHubUser,
-  Installation,
-} = models;
+const { SlackUser, GitHubUser, Installation } = models;
+
+const jestTimeout = 10000;
```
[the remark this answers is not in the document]
scepticulous: 5 seconds is the jest default. This PR doesn't change the value of our tests. 
I just extracted the value from the tests into a constant.

package.json
scepticulous: You are correct about the PATH, but we don't want to invoke jest, because that would use it's 'shebang' line which is running `node` without any flags. So that is, what we used before. 
The purpose of this change is to directly call the `node` executable and point it to the script it should run. It this case `jest`. So the PATH does only matter for the first command in the line (which is node). Hope that makes sense.

PR #896, opened 2019-08-05
Adds logging for skipped webhook events

scepticulous:
In order to understand how many webhook events we are discarding,
I added a simple logline, that should allow us to create filters
and dashboards for skipped and processed webhook events.

`/cc` @wilhelmklopp I am not 💯, if this covers all the cases. What do you think?

### DIFF
--- a/lib/activity/router.js
+++ b/lib/activity/router.js
@@ -20,8 +20,10 @@ module.exports = function route(callback) {
 
       const promise = Promise.all(subscriptions.map(async (subscription) => {
         if (!subscription.isEnabledForGitHubEvent(context.event)) {
+          context.log(`Checking subscription for event type: ${context.event}. (webhooks.subscription.match=0)`);
           return;
         }
+        context.log(`Checking subscription for event type: ${context.event}. (webhooks.subscription.match=1)`);
 
         const eventType = `${context.event}.${context.payload.action}`;
 


lib/activity/router.js
```diff
@@ -20,8 +20,10 @@ module.exports = function route(callback) {
 
       const promise = Promise.all(subscriptions.map(async (subscription) => {
         if (!subscription.isEnabledForGitHubEvent(context.event)) {
+          context.log(`Checking subscription for event type: ${context.event}. (webhooks.subscription.match=0)`);
           return;
         }
+        context.log(`Checking subscription for event type: ${context.event}. (webhooks.subscription.match=1)`);
```
wilhelmklopp: I think we can also pass a hash at the beginning. So could do something like:
```suggestion
        context.log({"webhook_subscription_math": true},`Checking subscription for event type: ${context.event}. (webhooks.subscription.match=1)`);  
```

But totally up to you 💯

lib/activity/router.js
```diff
@@ -20,8 +20,10 @@ module.exports = function route(callback) {
 
       const promise = Promise.all(subscriptions.map(async (subscription) => {
         if (!subscription.isEnabledForGitHubEvent(context.event)) {
+          context.log(`Checking subscription for event type: ${context.event}. (webhooks.subscription.match=0)`);
```
wilhelmklopp: We should double check whether we need to set an explicit log level here so that these will show up in Loggly. I don't remember exactly. 🤔

scepticulous: req.log (without passing the level) is in loggly:

https://github.com/integrations/slack/blob/e2ba204afb80619918929dd42804174374c0e579/lib/api.js#L51

`context.log.debug` (debug log level) is not.
https://github.com/integrations/slack/blob/e2ba204afb80619918929dd42804174374c0e579/lib/activity/router.js#L19

wilhelmklopp: Awesome! 

While we're here, let's also remove the `.debug` from line 19? I feel like we do definitely want to log that.

lib/activity/router.js
scepticulous: Awesome. Using a hash is soooo much better than coming up with cryptic unique key-value patterns 👍

scepticulous:
> Great work! Looking forward to see these figures.
> 
> I think we also want to have an event if `subscriptions` is an empty array: https://github.com/integrations/slack/pull/896/files#diff-b8f580f5fe58584eacf7fa5445166ed4R17
> 
> And in the case of org subscriptions where the creator doesn't have access: https://github.com/integrations/slack/pull/896/files#diff-b8f580f5fe58584eacf7fa5445166ed4R55

Thanks for pointing that out. 
You helped me understanding, that this PR is nonsense. 🤖 ..
I did not read the map iteration correctly. Thus this change would create a ton of unrelated log lines.
I tried to rewrite it in a way, that we cover all return statements, but only to realize, that this would also just create a line per subscription and not tell the action final result.

I don't see how to change this PR into something meaningful and think it's best to close it and go back to the drawing board.
I have an idea that involves changing the router a little bit more, but I will to that at a later time.

PR #908, opened 2019-08-14
Add Actions CI

scepticulous:
Run Actions CI in addition to Travis


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,39 @@
+name: GitHub CI
+
+on: [push, pr]
```
gimenete: I think the right event name is `pull_request` and right now is triggered in some circumstances in which you usually don’t want to trigger CI (e.g. a pr is deleted). There’s a synchronize event or similar. I’m on the phone so it’s hard to me to find an example now.

scepticulous: That makes sense. I think `push` is good enough anyways. Thanks for pointing that out 👍

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,41 @@
+name: GitHub CI
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x]
```
wilhelmklopp: Veeeery cool! ✨ 💪

PR #915, opened 2019-08-20
Use deployment-status ref instead of sha

### LINKED ISSUE #85: show ref instead of sha if one exists

scepticulous:
## Summary

When the integrator that created the deployment status provided a ref,
that is different from the commit sha, that ref is now displayed instead
of the short sha.
This addresses the issue discussed in #98 and #85.
I think the next step would be doing an actual API request to fetch the
branches, but his change is a 0-cost improvement and (almost) 0-effort
improvement that allows us to gather metrics about custom ref adoption
and maybe this already solves the problem.

## Screenshots

Before

depoyment-status-before

After

deployment-status-after

Closes #85, #98

### DIFF
--- a/lib/messages/deployment-status.js
+++ b/lib/messages/deployment-status.js
@@ -14,21 +14,17 @@ class DeploymentStatus extends Message {
     this.deploymentStatus = deploymentStatus;
     this.deployment = deployment;
     this.repository = repository;
+
+    const { ref, sha } = deployment;
+    this.shortSha = sha.substr(0, 7);
+    this.prettyRef = ref && ref !== sha ? `${ref} (${this.shortSha})` : this.shortSha;
     this.commitLink = `${this.repository.html_url}/commit/${this.deployment.sha}`;
   }
 
-  prettyRef() {
-    return this.deployment.ref && this.deployment.ref !== this.deployment.sha
-      ? this.deployment.ref
-      : this.deployment.sha.substr(0, 7);
-  }
-
   getCore() {
-    // @todo show ref instead of sha if one exists
-    // let center = `${this.shortSha}`;
-    let center = `${this.prettyRef()}`;
+    let center = `${this.prettyRef}`;
 
-    let centerWithLink = `<${this.commitLink}|\`${this.prettyRef()}\`>`;
+    let centerWithLink = `<${this.commitLink}|\`${this.prettyRef}\`>`;
     if (this.deployment.environment) {
       center += ` to ${this.deployment.environment}`;
       centerWithLink += this.deploymentStatus.target_url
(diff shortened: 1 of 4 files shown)


lib/messages/deployment-status.js
```diff
@@ -10,24 +10,30 @@ class DeploymentStatus extends Message {
...
   }
 
+  prettyRef() {
+    return this.deployment.ref && this.deployment.ref !== this.deployment.sha
+      ? this.deployment.ref
+      : this.deployment.sha.substr(0, 7);
+  }
+
   getCore() {
     // @todo show ref instead of sha if one exists
```
wilhelmklopp: We can finally get rid of this super old comment now 🙌

lib/messages/deployment-status.js
```diff
@@ -10,24 +10,30 @@ class DeploymentStatus extends Message {
...
-    let center = `${this.shortSha}`;
-    let centerWithLink = `<${this.commitLink}|\`${this.shortSha}\`>`;
+    // let center = `${this.shortSha}`;
```
wilhelmklopp: Should this line be sticking around?

lib/messages/deployment-status.js
```diff
@@ -10,24 +10,30 @@ class DeploymentStatus extends Message {
...
-
     });
     this.deploymentStatus = deploymentStatus;
     this.deployment = deployment;
     this.repository = repository;
-    this.shortSha = this.deployment.sha.substr(0, 7);
     this.commitLink = `${this.repository.html_url}/commit/${this.deployment.sha}`;
   }
 
+  prettyRef() {
```
wilhelmklopp: If we prefix this method definition with `get` then we don't need to always call it with `()` further down below.

```suggestion
  get prettyRef() {
```

Prior art: https://github.com/integrations/slack/blob/fad9c6409bf720b3b36bae3b1e15ad73dc7fdfc4/lib/messages/flow/updated-settings.js#L10-L28

Totally just an idea though, feel free to ignore

lib/messages/deployment-status.js
```diff
@@ -10,24 +10,30 @@ class DeploymentStatus extends Message {
...
+    return this.deployment.ref && this.deployment.ref !== this.deployment.sha
+      ? this.deployment.ref
```
wilhelmklopp: Does it make sense to include the sha alongside the ref if we do have one? So something like:

```suggestion
      ? `${this.deployment.ref} (${this.deployment.sha.substr(0, 7);})`
```

So that in Slack it shows up as 
"login is deploying `master` (abc1def2) to staging"

scepticulous: That's a great idea 👏 
I Added it, added screenshot and implemented your other suggestions too.

PR #916, opened 2019-08-21
Production package updates (minor only)

scepticulous:
## Summary

Perform minor package update for production packages. 
The updates should be safe to do, but since they affect production, I separated them from the dev package update.

We also had to adjust some of the tests to remove `async` from `describe` statements. Otherwise, `jest` would no longer run these tests.

Finally, I had to update older migrations that had more than one statement because the minor sequelize update broke our migrations in CI, because the new version fails when the migration function does not return a promise.
See https://github.com/sequelize/umzug/issues/185 for all details.


db/migrations/20171207225726-create-slack-workspace.js
```diff
@@ -1,6 +1,7 @@
 
 module.exports = {
   up: (queryInterface, Sequelize) => {
+    return Promise.all([
```
wilhelmklopp: I assume the sequelize version change required these migration changes? Weird that they make a breaking change for a minor version bump, but maybe we were "holding it wrong" the entire time 😄

scepticulous: Yes, exactly. The minor change caused our migrations to fail in CI. I think I mentioned this in one of my commits but forgot to mention in the in PR body.

PR #920, opened 2019-08-22
Increase jest timeouts to stablize unfurl CI tests

scepticulous:
Actions CI tests run into the jest default timeout of 5 seconds on
almost all builds. This increases the value to 10 seconds, similarly to
other integration tests we have.

GitHub CI is ✅ . It works 👏

### DIFF
--- a/test/integration/unfurl.test.js
+++ b/test/integration/unfurl.test.js
@@ -30,6 +30,10 @@ describe('Integration: unfurls', () => {
   });
 
   describe('public unfurls', () => {
+    beforeEach(async () => {
+      jest.setTimeout(10000);
+    });
+
     test('issue', async () => {
       nock('https://api.github.com').get('/repos/bkeepers/dotenv').times(2).reply(
         200,


jules2689:
> Actions CI tests run into the jest default timeout of 5 seconds on almost all builds. 

What do you mean by this? How does Actions CI run into a timeout of what it's running?

scepticulous:
> > Actions CI tests run into the jest default timeout of 5 seconds on almost all builds.
> 
> What do you mean by this? How does Actions CI run into a timeout of what it's running?

The integration tests seem to be affected by the hardware performance of the system running then. Even though all requests are (should be) stubbed, the slower the system, the longer it takes our tests to be executed(`nock`-mocks, redis-calls, database-calls and probot logic). 
This is even observable on the local system. The slower your system gets, the higher the frequency of failing tests, due to exceeding the `jestTimeout`. 
Travis is a little slower than our new MacBooks and because of that we sometimes run into jest timeouts on Travis but only very rarely on local setups. 
GitHub CI executes slower than Travis and fails even more frequently with `jestTimeout` errors. I am not really sure, why we are running into these errors in the first place, but understanding and fixing this might take some time.

PR #929, opened 2019-08-30
Fix issue comment unfurl

### LINKED ISSUE #928: Unfurl issue comments
# Unfurling issue comments causes errors

## Summary

Sharing a link to a GitHub issue comment leads to an error message.

### Visibility
For public repos no message is shown
For private repos an error message is shown, asking the user to contact GH support.

### Expected behavior

The GH issue comment is unfurled

### Steps to reproduce
* Share an issue comment link on GitHub for a private or public channel
* Link is not unfurled 
* For private repositories and error is shown to the user.

Example link: 
* https://github.com/rust-lang/rust/issues/63997#issuecomment-526118914

### Cause

```
8:56:48.513Z ERROR http: Empty value for parameter 'comment_id': undefined (id=86b1ba8d-141f-40e9-9eef-ebc87e5f0be3)
  HttpError: Empty value for parameter 'comment_id': undefined
      at values.forEach (/Users/.../slack/node_modules/@octokit/rest/plugins/validate/validate.js:66:15)
      at Array.forEach (<anonymous>)
      at Object.keys.forEach.parameterName (/Users/.../slack/node_modules/@octokit/rest/plugins/validate/validate.js:38:12)
      at Array.forEach (<anonymous>)
      at validate (/Users/...s/slack/node_modules/@octokit/rest/plugins/validate/validate.js:15:23)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

scepticulous:
Closes #928

## Problem
This commit fixes unfurling of private and public issue comments.
When the `octokit/rest.js` changed [their routing arguments](https://github.com/octokit/rest.js/commit/c62b2bcb2c4f66d93ecff1563df7f7d8e8431788) from `id` to `<resource>_id`, unfurling issue comments stopped working.

## Solution

This PR adjusts the argument name and thus fixes the problem.

## Example (from linked issue)

**URL**: `https://github.com/rust-lang/rust/issues/63997#issuecomment-526118914`
**Error**:
```
8:56:48.513Z ERROR http: Empty value for parameter 'comment_id': undefined (id=86b1ba8d-141f-40e9-9eef-ebc87e5f0be3)
  HttpError: Empty value for parameter 'comment_id': undefined
      at values.forEach (/Users/.../slack/node_modules/@octokit/rest/plugins/validate/validate.js:66:15)
      at Array.forEach (<anonymous>)
      at Object.keys.forEach.parameterName (/Users/.../slack/node_modules/@octokit/rest/plugins/validate/validate.js:38:12)
      at Array.forEach (<anonymous>)
      at validate (/Users/...s/slack/node_modules/@octokit/rest/plugins/validate/validate.js:15:23)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```
**Offending code**:
We are passing `id: id` (shorthand version) instead of `comment_id: id`.

https://github.com/integrations/slack/blob/8aa8e7b8763a404db96781d114abd8f3f58e8bbd/lib/unfurls/comment.js#L8-L13

## Additional fixtures

In addition, I added new fixtures to allow tests that have less context.
This should make it clear, that for public unfurls, the user of the repo
and the comment does not matter and it decouples the fixture and
subscription setup.
This was mainly helpful to me to make sure the tests only require, what they should require.

ℹ️ These fixtures are very similar to the existing ones and can be ignored during the review if that helps.


lib/unfurls/index.js
```diff
@@ -56,10 +56,11 @@ async function linkShared(req, res) {
       });
     } catch (err) {
       if (err.name === 'UnsupportedResource' || err.code === 404) {
-        req.log.debug(err, 'Could not get unfurl attachment');
+        req.log.warn(err, 'Could not get unfurl attachment');
```
wilhelmklopp: I _think_ `UnsupportedResource` resource might happen quite a lot because there are so many github links we don't have unfurls for, so it's possible that this will blow up the logs.

But happy to get this out, see if that's actually the case and then readjust to necessary. Ship to learn 🚀

scepticulous: Good catch ❗️ 
Initially, I was thinking the same but I was unable to estimate how common that case is.
The better approach seems to be logging it as 'debug' and then we can control it via Heroku `LOG_LEVEL` if we want to take a look at the most common URLs.

scepticulous: I changed it back to `debug` in https://github.com/integrations/slack/pull/929/commits/dce6de270308e4e5089470a18d992f631b3785b1

gr2m:
> I wonder how we can spot something like this earlier when we next upgrade Probot or Octokit. Should Octokit have thrown errors/warnings when passing old arguments that no longer work? cc @gr2m

Ocotkit is logging warnings. The commit you linked shows that the previous URL parameter `:id` was deprecated and was aliased to `:thread_id`, you should have seen a deprecation warning if you still used `id` instead of `thread_id`. 

For breaking versions I usually try to only remove previously deprecated changes. So make sure to update to the latest version of the previous major version, run your tests, address all deprecations, then upgrading should be simple.

PR #936, opened 2019-09-13
Add Indexes on Unfurls

### LINKED ISSUE #869: Add DB indexes on unfurls table to improve debuggability
Errors with unfurls can be hard to debug at the moment, in part because the database table is quite large and because we don't have indexes on any of the columns that are often used for debugging, digging into a particular issue.

Add indexes 
- [ ]  chanelSlackId
- [ ]  slackWorkspaceId
- [ ] slackUserId
- [ ] githubRepoId

srinjoym:
Fixes #869 

Adds indexes on the `channelSlackId`, `slackWorkspaceId`, `slackUserId` and `githubRepoId` columns on the Unfurls table.

### DIFF
--- a/db/migrations/20190913165643-add-indexes-to-unfurls-table.js
+++ b/db/migrations/20190913165643-add-indexes-to-unfurls-table.js
@@ -1,11 +1,26 @@
 module.exports = {
   up: (queryInterface) => Promise.all([
     queryInterface.addIndex('Unfurls', {
-      fields: ['channelSlackId', 'slackWorkspaceId', 'slackUserId', 'githubRepoId'],
-      name: 'unfurls_slack_channel_workspace_user_ids_github_repo_id_idx',
+      fields: ['channelSlackId'],
+      name: 'unfurls_channel_slack_id',
+    }),
+    queryInterface.addIndex('Unfurls', {
+      fields: ['slackWorkspaceId'],
+      name: 'unfurls_slack_workspace_id',
+    }),
+    queryInterface.addIndex('Unfurls', {
+      fields: ['slackUserId'],
+      name: 'unfurls_slack_user_id',
+    }),
+    queryInterface.addIndex('Unfurls', {
+      fields: ['githubRepoId'],
+      name: 'unfurls_github_repo_id',
     }),
   ]),
   down: (queryInterface) => Promise.all([
-    queryInterface.removeIndex('Unfurls', 'unfurls_channel_slack_id_slack_workspace_id_slack_user_id_github_repo_id'),
+    queryInterface.removeIndex('Unfurls', 'unfurls_channel_slack_id'),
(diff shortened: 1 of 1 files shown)


db/migrations/20190913165643-add-indexes-to-unfurls-table.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  up: (queryInterface) => Promise.all([
+    queryInterface.addIndex('Unfurls', {
+      fields: ['channelSlackId', 'slackWorkspaceId', 'slackUserId', 'githubRepoId'],
+      name: 'unfurls_channel_slack_id_slack_workspace_id_slack_user_id_github_repo_id',
```
jules2689: These names are a bit long. Postgres has an identifier limit of 64, so the max # of characters is 63.
This is 72 characters, and will be truncated.

https://www.postgresql.org/docs/9.1/sql-syntax-lexical.html

This might work:

```suggestion
      name: 'unfurls_slack_channel_workspace_user_ids_github_repo_id_idx',
```

db/migrations/20190913165643-add-indexes-to-unfurls-table.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  up: (queryInterface) => Promise.all([
+    queryInterface.addIndex('Unfurls', {
+      fields: ['channelSlackId', 'slackWorkspaceId', 'slackUserId', 'githubRepoId'],
```
wilhelmklopp: Awesome, thanks for getting this started ✨ 

I'm not 100% sure here, but if we add the index this way, is it a "composite" index that only ends up becoming helpful if we have a query that specifies all four fields in the `WHERE` clause?

I've previously seen indexes get added individually for each column, like in this migration: https://github.com/integrations/slack/blob/7d4c2d323396c06e7744665b5e008f241154d2f5/db/migrations/20180117192603-add-indexes.js#L4-L23

jnraine: @wilhelmklopp is exactly right. 

> A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but **the index is most efficient when there are constraints on the leading (leftmost) columns.** The exact rule is that equality constraints on leading columns, plus any inequality constraints on the first column that does not have an equality constraint, will be used to limit the portion of the index that is scanned. 
> – https://www.postgresql.org/docs/10/indexes-multicolumn.html

Unless we plan to query against all four columns in the same query or only the leftmost columns, splitting this into four, one-column indexes will provide the most flexibility while debugging.

I'm happy to chat about this further. Some of these column constraints can be difficult to explain without examples.

srinjoym:
Thanks for the feedback everyone! Sorry for the delay getting back to this but I switched this to be 4 one column indexes! 🎉

tebriel:
Shelving these until we find that we need them.

PR #953, opened 2019-10-10
Show more appropriate icons for pull requests

zlatkoverk:
It is confusing when new pull request is created and it says No checks have passed with green check mark.

https://prnt.sc/phcc7u


tebriel:
Thanks @zlatkoverk, any reason why you changed the default from ✅ to ☑️ ?

zlatkoverk:
So that we make a distinction between `pending` status and `success` status

Issue #984, opened 2020-01-29
[Feature Request] Deployment parameters / errors 


iboobel:
We are trying to use deployments via slack. 
We faced a few issues: 
- can't set additional deployment parameters (like auto_merge or required_contexts  https://developer.github.com/v3/repos/deployments/#create-a-deployment)
-  can't see why deploy wasn't run. E.g.:  if the status check is false, the deploy will not run and slack doesn't show the reason

tebriel:
Thanks @iboobel. Deploys are one of our least-used features (we see less than 100 a day could be a chicken-egg situation where they're not what people need so they're not used) so they've been lagging in feature development. I don't think in the near-future we'll be able to add this requested functionality due to our other priorities (changing the infrastructure the app is deployed on and fixing some long-standing bugs). I very much like this idea, though.

mkrakowitzer:
Would also like this feature built into the slack integration. 

It's listed in the help as a feature, but it just errors out and it's undocumented if it accepts any parameters at all such as environment etc.

```
/github deploy owner/repository
```
@iboobel it looks like you have limited success with the deployments, if you could give me some pointers on how it to use them would appreciate it.

tebriel:
@mkrakowitzer when you issue a `/github deploy owner/repo` it will (should, except in the case of an error) open up a Slack Dialog to allow you to pick some default options. Screenshot attached, sorry it isn't working for you.
![Deployment Dialog](https://user-images.githubusercontent.com/821688/73672487-f21a8480-467a-11ea-907a-aa4846b35741.png)

shandyDev:
It actually strange: payload (very optional) have in dialog popup, but another 2 options (https://developer.github.com/v3/repos/deployments/#parameters-1):
- auto_merge
- required_contexts

It very important params because deployment (special in CI processes) should be start as is without action on codebase (merge) also without CI checks.
What all need IMHO, add fields to dialog and pass it to github api:
https://github.com/integrations/slack/blob/master/lib/messages/create/create-deployment-dialog.js#L58
https://github.com/integrations/slack/blob/master/lib/create/deployment.js#L38

PR #995, opened 2020-02-19
Deployment improvements

### LINKED ISSUE #975: "Trigger a deployment" dialog includes incomplete set of branches and tags
If I run `/github deploy org/repo` command I'm presenter with "Trigger a deployment" dialog. The problem that the "Branch or tag to deploy" select box has an incomplete list of branches and tags, it's probably due to not paginated requests here:
https://github.com/integrations/slack/blob/09969eb928c6e323e15311129698c75e5c6c2c66/lib/create/deployment.js#L21-L22
so we get only first 30 items by default.

I believe this is a bug as we should be able to deploy any branch or tag from the repository.

alexjurkiewicz:
* Allow freeform `ref` input (rather than 100-item-max select input)
* Add `auto_merge`
* Add `description`

Fixes #975
Some work on #984


alexjurkiewicz:
CI tests seem to be failing due to some sort of integration snapshot comparing before and after. I'm not very familiar with the tools in use -- any help on how to update the test snapshots?

kdrakon:
@alexjurkiewicz looking forward to your improvement (exactly the change we were looking for). Are you ok with @kawabatas' test fix?

alexjurkiewicz:
Thanks for the prompt! I missed that PR 🙏

alexjurkiewicz:
It looks like there are still CI issues :(
We implemented another solution for triggering deployments -- if someone else wants to take this PR over I will close it.

kdrakon:
@alexjurkiewicz thanks for retrying the PR. Can you let me know what your other solution was for triggering deployments, in the meantime? I presume you are hitting the same 'auto-merge' issue we are (i.e. we want to be able to deploy past branches without a merge from master).

kawabatas:
@alexjurkiewicz 
>It looks like there are still CI issues :(

I think so, too.

Can you retry this failed travis build ?
I think CI will pass if you retry a few times.

alexjurkiewicz:
> Can you let me know what your other solution was for triggering deployments, in the meantime? I presume you are hitting the same 'auto-merge' issue we are (i.e. we want to be able to deploy past branches without a merge from master).

Sorry, we don't hit that issue -- our build system produces artifacts so it sounds like our workflow is a little different. We've changed to triggering deployments via a PR comment + github actions.

eliothxchan:
Any news on getting this merged? This would be super helpful to my team!!

kdrakon:
Would love to get this improvement too please 🙏

varevarao-airmeet:
+1 Is this being picked up soon? This is a very useful feature we'd like to get running in our workflows, and is only limited by the incomplete refs list.

rikoe:
+1, as I just raised #1030 to disable auto-merge, as it is really unhelpful (I would argue that having auto_merge=true by default on the event is the real problem, I can't imagine when you would want that on by default)

When can this be merged, all the checks are passing?

lib/messages/create/create-deployment-dialog.js
```diff
@@ -37,10 +37,11 @@ module.exports = ({ repository, branches, tags }) => {
         }],
       },
       {
-        label: 'Branch or tag to deploy',
-        type: 'select',
+        type: 'text',
+        label: 'Branch/tag/sha to deploy',
```
rikoe, line 41: I think changing this from a select to a text control makes it less useful, it is of great benefit to just be able to select an existing branch or tag from a list.

Having to type refs in by hand will be much more error-prone.

kdrakon: But a select drop-down does not scale with the number of potential branches and tags a repository may have. In our case, we tag every build, and because of the current integrations limitation of 100 select items, we are forced to delete older tagged releases (something I'd rather we didn't do). Free-text input is error-prone I agree, but I personally would prefer this trade-off.

rikoe: I wouldn't prefer this option though. I think the user experience is a lot worse, for the sake of a small number of users who have a lot of tags/branches. Is there a compromise? Can we support some combination of select and free-form text entry?

alexjurkiewicz: I'm open to your suggestion. Please read the slack API docs and get back to us.

rikoe: How about the Slack external source select element, combined with the a GitHub ref search?

https://api.slack.com/reference/block-kit/block-elements#external_select
https://developer.github.com/v3/git/refs/#list-matching-references

I don't know the Slack or GitHub APIs very well, so what I suggest may not be possible, but at first glance both APIs support searches based on partial matches of text.

I think going from a select field to a free-form entry field would be a backward step, and I would encourage looking at all possible solutions. At the moment, non-developers can trigger deployments via Slack, as they do not have to have intimate knowledge of tags, branches etc. The select dialog helps them to choose by displaying known branches and tags.

Two other options I can think of in case the above is not workable:
* Having both the current select control, and the text box, with the one being populated taking precedence.
* Having the option to configure which type of control (select or text) to use in the Slack app settings.

bouwerp:
Any traction on this? Can I help, perhaps? We're busy building a custom GitHub app to manage deployments, but if we can do this through Slack, we'll put that on the ice. We can't access out latest tags, so we really can't use it before then.

lib/messages/create/create-deployment-dialog.js
rikoe: Please ignore my comments above - let's get this merged so people can use this integration. One can always iterate on improvements for selecting tags/branches later.

achempak:
Any news on merging this? It would be super useful for my team :pray:. The alternative is building a custom app to trigger deployments, but would rather not do that if possible.

ric2z:
> Any news on merging this? It would be super useful for my team 🙏. The alternative is building a custom app to trigger deployments, but would rather not do that if possible.

Same here, this would be very useful for our team 🙏

nitrocode:
Yes, please. Please review the code @jules2689 when time permits. Thank you.

pmalon:
any chances for getting it merged? Those changes solve real problems that GH Slack integration users have on a daily basis...

pmalon (approved):

ashokirla:
Hi, Thanks for your contribution. We are going through platform changes to our ChatOps app and are not accepting any external contributions.  We will go through your PR and consider your suggestions if it aligns with our Product direction. We are converting your PR to a feature request and it can be tracked [here](https://github.com/integrations/slack/issues/1246).

Issue #767, opened 2019-01-18
Migrate away from being a Slack "Workspace App"


abinoda:
It appears that contributing to this integration is not possible right now because per the [Contributing Docs](https://github.com/abinoda/slack/blob/master/CONTRIBUTING.md#getting-started), local development requires a Slack Workspace app, which can no longer be created.

I'm aware that there are plans to do this migration in the future but I'm wondering if there's a short-term workaround so local development is possible?

gimenete:
Sorry for that! We didn't expect this. We haven't looked into the migration path yet. We will keep you up to date about our progress as soon as we start working on that.

Shegox:
I just run into this issue as well, as I wanted to experiment with GitHub Enterprise support. 
Are there any plans to migrate it away from a workspace app yet?

scepticulous:
Hi everybody 👋 ,

we are working with slack on migrating the Slack integration from workspace tokens to a different solution.  We do not have an ETA yet, but we will share updates here.

kavdev:
@scepticulous thanks for the update! Any workarounds for people looking to test locally to make contributions would also be much appreciated

Issue #565, opened 2018-05-03
GitHub Enterprise support


bkeepers:
From the README:

> This app officially supports GitHub.com and Slack.com but the team plans to support GitHub Enterprise and Slack Enterprise Grid in the future.

Our approach to GitHub enterprise support will be two-fold:

1. Allow customers to run their own instance of the integration alongside their GitHub Enterprise installation

2. Add support for GitHub Enterprise to the hosted integration

We have not current prioritized this work, but will consider it based on demand. Add a 👍 reaction to this issue to show your support.

madsid:
Thanks for making this bot. 
Just checking if there are any updates on this? I am debating either to wait for the update or use an alternative ( we host our own enterprise GitHub)

bkeepers:
@madsid, we're not actively working on this, but I reopening the issue since this is a feature that many others would like.

jeffj254:
This would definitely be very useful for those of us that run internal GHE but use Slack.com

davesade:
Add me to the waiting list or let us know, how this can be done otherwise. We are happy users of Slack and GitHub Enterprise, so it absolutely makes sense to allow users to choose their origin service, instead of static github.com. Afaik previous GitHub integration on Slack allowed to connect GHEE repos to it (iirc it was just a webhook between repo and Slack), so for many of us this is a must have feature. Thanks a lot in advance!

seangates:
Bump!!

gpaulsen:
+1 vote for Enterprise integration.

ponchieponcho:
Would love for this feature to get attention. Posting github links in Slack doesn't format them at all until this is finished. I started building a Slack app that hits the Github API and I format the PRs myself at the moment.

seangates:
Is there any progress on this integration? There are quite a few of us patiently waiting to empower our teams. Especially since some of us have thousands of developers who would benefit form it's capabilities.

tcbyrd:
Since we built the integration as a workspace app and Slack is moving those features to bot tokens, it means the app won't work for Enterprise customers in its current form (you wouldn't be able to create the necessary Slack App). There's some refactoring needed to make this work with bot tokens, but then it should be much easier to setup. We don't have a timeline for completing this, but also if anyone wants to dive into the code and help out with that effort, feel free.

kconway-tangerine:
Hello:

[...]

PR #1055, opened 2020-07-31
Allow self-hosted for GHE and Slack Grid

barlock:
**Related Issues**
#565 #767

**What does this PR do?**  
Allows for the app to be self-hosted supporting Github Enterprise (server 😉) and a specific Enterprise Grid instance.

- Removes hard coded references to github.com and allows the `probot` env variable `GHE_HOST` 
- Allows for self signed certs to be passed in for redis and postgres. (I don't understand why, but that's the only option I have in my hosted environment)
- Upgrades OAuth from workspace tokens to `v2`. 
  - Needed for new apps as you can't create "Workspace" apps anymore
  - @aoberoi (DevRel from Slack) is going to help review this to make sure that oAuth doesn't break on the "production" instance. 
- Lets you customize a `SLACK_ENTERPRISE_ID` that allow allows installs for workspaces in that account
  - Prevents information leakage from url unfurls. 

**What gif most accurately describes how I feel towards this PR?**


aoberoi:
I'm excited about this! It's going to take me a few days to conduct a review, but I'm happy to help.

Just so we're clear on expectations for not breaking production, we're aiming for the following:
* for existing deployments, installations in Slack with a Workspace token (`xoxa`), should continue working without any user-facing changes. simply updating the deployment to this version will preserve all existing functionality in those Workspaces.
* for existing deployments, all new installations will occur using the OAuth 2.0 v2 (aka new Slack apps, aka granular bot permissions) and result in an `xoxb` token.
* a single deployment can support both installations with a workspace token (`xoxa`) and installations using OAuth 2.0 v2 (`xoxb`) at the same time.

☝️ That's all requirements I just made up based on my understanding of how this repo is used by the deployment which supports the Slack App Directory, and will now be expected to work for on-prem or self-hosted deployments. But please correct me if I'm wrong!

**Question**: Does this PR already, or should it, add a user-facing means to ask a user to "upgrade" their installation (transition from using an `xoxa` to an `xoxb`)? If we made this a requirement, we could gradually start reducing the number of Workspace App installations that exist in the wild. We'd have to carefully consider which users, how, and when they should be prompted to "upgrade".

barlock:
> Does this PR already, or should it, add a user-facing means to ask a user to "upgrade" their installation.

No it doesn't. I think that would live better in a separate pr. I also noticed that the app still uses `verification tokens` instead of the newer `signing secret`, which would fit well in another PR too.

bar:
@barlock do you know why the changes have not been approved? I understand the GH team have plans to support GHE server and grid which might be already in the works?

barlock:
Nope, not sure why this PR hasn't gotten any attention. 

I've got it deployed within my Company and it's been working pretty well. There are a few things where the App's UX isn't super great for non-workspace tokens, but it's pretty good for the effort.

yaustar:
Bump! Would love to see this integrated for enterprise support too.

redrac:
Bumping as well! We would love to use this at our company with grid and GHE server as well!

hmc-cs-mdrissi:
One issue that appears to be true is no one from github is a participant in this pr. The one person you @aoberoi isn't at github anymore so would be unlikely to review. Some people that could maybe review, @julian-alarcon or @ashokirla . Looking at a comment from another pr, 

> @jmfrancois Thanks for your contribution. Unfortunately, we are not taking contributions at this point due to few platform changes. We will move this request to our backlog and take it up separately.

I'm unsure as to the state of this repository for new features.

ashokirla:
Hi, Thanks for your contribution. We are going through platform changes to our ChatOps app and are not accepting any external contributions. However, we are currently working on adding support for GHES and Enterprise grid. We will look into your suggestions as part of it.

barlock:
🙇 Keep me posted and if you open source it drop a link! I'd love to contribute. 

I'd also love to shut down the instance I run which is _very_ buggy as I obviously can't get workspace tokens anymore.

PR #1064, opened 2020-08-31
More inclusive language

mattwynne:
Changes "whitelist" to "allow-list"

I know it's not going to save the world but it also can't do any harm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ And to turn it back off:
 
 #### Filters
 
-Label filters allow filtering incoming events based on a whitelist of **required** labels.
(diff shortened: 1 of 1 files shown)


sandstrom:
There is nothing exclusive about the word white or white list, just as the word red or green isn't. They are just words.

https://en.wikipedia.org/wiki/Blacklisting#Origins_of_the_term

cgriego:
It should just be "list of **required** labels", that's the clearest language.

ashokirla:
Hi, Thanks for your contribution. We are going through platform changes to our ChatOps app and are not accepting any external contributions.  We will consider your suggestion as part of the doc update.

Issue #970, opened 2019-12-11
[Bug] Labels with "_" are not valid arguments


StummeJ:
This is a pretty cool slack plugin. I've however noticed when trying to filter the subscription by a label with an underscore in it, I get an error `+label:something_with_underscores is not a valid argument`.  This happens when it's quoted (double or single) or just raw.

Cheers

rmacklin:
I can confirm that this issue is still present as of March 5, 2020

tzahifurmanski:
I can confirm that this issue is still present as of August 13, 2020

joeydong:
+1

cpoonolly:
+1 Think it's failing for other characters as well:
https://github.com/integrations/slack/issues/958

soudai:
+1
I can confirm that this issue is still present as of January,19 2021

razajafri:
+1 This is still a valid issue I was trying it with a label that has paranthesis

Piet-G:
+1 Still valid as of 10 December 2021.

Very frustrating, the error message is unclear as well. According to #1076 the fix should be a simple addition to a config file

fetis:
2022 and seems it's still the case

rcheemaupgrade:
2022 Aug - We are facing same issue.

cagesellchen:
👋 This has now been fixed!

PR #1076, opened 2020-10-13
Allow underscore label

t-cyrill:
Related issue: https://github.com/integrations/slack/issues/970

I get an error +label:something_with_underscores is not a valid argument, when trying to filter the subscription by a label with an underscore.

I read "Valid filters" section in README, there is no mention of underscore.

Then I checked `labelPattern` regex rule in `settings-helper.js`, underscore is not allowed in label pattern.

Is it possible to add an underscore to it?

Thank you for reading.

### DIFF
--- a/lib/settings-helper.js
+++ b/lib/settings-helper.js
@@ -13,7 +13,7 @@ function labelValuePattern() {
 // label:foo label:teams/designers label:"foo bar" label:' bar baz'
 
 function labelPattern() {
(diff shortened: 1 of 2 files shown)


ashokirla:
Hi, Thanks for your contribution. We are going through platform changes to our ChatOps app and are not accepting any external contributions.  We will go through your PR and consider your suggestions if it aligns with our Product direction. We are converting your PR to a feature request and it can be tracked [here](https://github.com/integrations/slack/issues/1248).

PR #1130, opened 2021-02-25
fix(release): fallback to tag_name when release name is null

jmfrancois:
We have often this kind of message:

Screen Shot 2021-02-25 at 10 06 36

This is because the user do not fill the release name. IN the github UI display the tag name in this case.
Align slack integration with this

### DIFF
--- a/lib/messages/release.js
+++ b/lib/messages/release.js
@@ -17,11 +17,12 @@ class Release extends Message {
   getRenderedMessage() {
     const { release, repository } = this;
     const { author } = release;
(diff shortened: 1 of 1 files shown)


ashokirla:
@jmfrancois Thanks for your contribution. Unfortunately, we are not taking contributions at this point due to few platform changes. We will move this request to our backlog and take it up separately.

ashokirla:
Moving this request to backlog [here](https://github.com/integrations/slack/issues/1249).

Issue #774, opened 2019-01-23
GitHub events cause Subscriptions to be removed


kpollich:
Hello,

We're having issues with our GitHub integration in one particular private Slack channel. We have dozens of other private channels that are working perfectly on the non-legacy GitHub integration.

In this channel, any subscribed GitHub events that fire cause the subscription to be deleted. So, subscribing to comments on PR's and then commenting on a PR causes the subscription to that repo to be removed.

I'm attaching a gif showing this behavior. I've also created a one-off channel to verify that we can receive events successfully from these repos, and that channel is working correctly.

https://giant.gfycat.com/DaringMiniatureBluefish.webm

The gif shows the workflow for reproducing the bug in the channel
1. Subscribe to new repo
2. Comment on a PR in that repo
3. Subscription has been removed from Slack channel

joshocalico:
|![Issues](https://user-images.githubusercontent.com/35418916/52255531-df377c00-2962-11e9-8b39-b06ab0057b38.png)|
|:---:|
|_Excuse my crudely drawn redaction lines_ |

I'm having the same issue when I create open issues through the integration. That itself works fine but as you can see from the image after creation the list gets wiped. It doesn't even announce the issue.

To reproduce this particular case - 

1. Add GitHub to slack. 
2. Subscribe to a repository on a private channel.
3. Use "/github subscribe list" for good measure.
4. Use "/github open [Your User]/[Your Repo]
5. List the subscribers again.
6. Weep that nobody knows about your lovely fresh Issue.

Something is definitely up here, likely specific to private repositories anyway.

asaletnik:
Same here:
- configured 2 private slack channels to track one private repository, both the same way
- event happens, channel 1 displays the notification, channel 2 loses the subscription

The one that fails was the first channel where I tried configuring github integration, it never worked. Tried removing the app and re-inviting it to that channel then setting up again with no luck. Worked on the other channel without any issues at first try.

byoung592:
I had the same issue! 

I got it to work by adding the integration/app to the whole workspace and then just subscribed my repository with my channel.

skylineproject:
I was originally running into this issue when connecting a private repo with a shared channel. I had added the app to the specific shared channel I was using, and, as described above, the subscription would seem to work, and then after some time passed, the subscription list command would return an empty list again.

After following @byoung592's advice, I'm now running into a new issue (#650) – where I can't even add the app to this channel. Is there any movement on this issue where the app silently unsubscribes?

[...]

PR #1146, opened 2021-04-09
Add note that private Slack channels are no longer supported.

rynmlng:
After contacting GitHub and Slack support they have reported back that private channels will no longer receive notifications.

Private channels integrated before this bug was introduced may still work, but are not guaranteed to continue working.

[See this issue and conversation w/GitHub and Slack](https://github.com/integrations/slack/issues/774#issuecomment-816751146)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ This app officially supports GitHub.com (which includes our GitHub Enterprise cl
 
 <p align="center"><img width="450" alt="auth" src="https://user-images.githubusercontent.com/3877742/36522927-f1d596b6-1753-11e8-9f85-2495e657b16b.png"></p>
 
+**Note:** At this time this integration only works with public Slack channels. While a private Slack channel can subscribe to a repository, it will not post notifications.
+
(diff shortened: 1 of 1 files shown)


julian-alarcon (approved):
LGTM

README.md
```diff
@@ -34,6 +34,8 @@ This app officially supports GitHub.com (which includes our GitHub Enterprise cl
 
 <p align="center"><img width="450" alt="auth" src="https://user-images.githubusercontent.com/3877742/36522927-f1d596b6-1753-11e8-9f85-2495e657b16b.png"></p>
 
+**Note:** At this time this integration only works with public Slack channels. While a private Slack channel can subscribe to a repository, it will not post notifications.
```
amitkumariiit, line 37: This is not the case. App works for the private channels too. To get it working app must be invited to the private channel.
Also thing to note is that, when installing the app, it should not be installed only on a specific channel. If it's installed for all channels and app is invited in the channel, it must work.

rynmlng: Then it sounds like the best solution is to update this note and say "app will not work on private channels unless app has access to all channels in initial integration?"

amitkumariiit: This point is already mentioned in the next line. (line no 39) :)

PR #1179, opened 2021-04-20
Update README.md to point to new Github app in the slack marketplace

sushmita-m:
Updated the GitHub app link to the market place in the readme file.


apurva1112 (approved):
Will be good to merge in case it does not regress anything.

PR #1220, opened 2021-06-03
Update README.md

ashokirla:
Added deployment notifications


masayyad (approved):
looks good

ocean90:
It would have been nice if the new default would only affect new subscriptions and not existing subscriptions. It was a bit surprising to suddenly see these notifications while never subscribed to those.

PR #1227, opened 2021-06-07
Fix signout suggestions

jasminesundaymatcha:
test/fixtures/slack/config_migration_single.json


amitkumariiit:
@jasminesundaymatcha we are not accepting the external contributions for now and planning to remove the code from this repo soon. https://github.com/integrations/slack/pull/1226

Please open an issue for any concern or issue in the signout behaviour and we should take care of.

PR #1443, opened 2022-08-17
[GHES] Added manifest for Slack app

aasiddiq:
Added the manifest for slack for app creation to support GHES


ashokirla (changes requested):
@aasiddiq Teams manifest should be in integrations/ms-teams repo. We will also breakdown the steps for each of them separate and put them in both repos respectively.

aasiddiq:
> @aasiddiq Teams manifest should be in integrations/ms-teams repo. We will also breakdown the steps for each of them separate and put them in both repos respectively.

Removed the team manifest and added in https://github.com/integrations/microsoft-teams/pull/226

ashokirla (approved):

PR #1514, opened 2022-10-28
Update README.md

bake0937:
Modify the README so that the command can be executed because the following command causes a Validation Error

Before

```bash
/github subscribe org/repo comments:"channel"
```
image

After

```bash
/github subscribe org/repo comments:“channel“
```


gr2m:
cannot reproduce

PR #1805, opened 2023-11-20
Add signin example to mentions section

### LINKED ISSUE #1382: Docs for mention mapping are unclear
From the README:
> Mentions will work only if you login to GitHub app in your Slack workspace. When you login to GitHub app with your GitHub id, we map it with your Slack id and ping you in Slack whenever you are mentioned in any of the GitHub notifications.

What does that mean exactly? How does an ordinary user "log into their GitHub app in their Slack workspace"? What instructions should we give our users to set up mentions?

haydn:
The mentions section explains that you need to login to the GitHub app in Slack, but doesn't explain how to do that. The `signin` command it only mentioned in once other place in the entire README and it's very easy to miss it.

Closes #1382.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ And the best part is - you can now see the summary of GitHub notifications where
 
 
 #### How does mentions work?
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Thank you!

PR #1943, opened 2024-09-05
Update README.md

O5ten:
I believe that users were the intended pattern here and not branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Branch filters allow filtering commit notifications. By default when you subscri
 - `/github subscribe org/repo commits`  for commit notifications from a default branch.
 - `/github subscribe org/repo commits:*`  for commit notifications across all the branches.
 - `/github subscribe org/repo commits:myBranch`  for commit notifications from a specific branch.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,7 +138,7 @@ Branch filters allow filtering commit notifications. By default when you subscri
 - `/github subscribe org/repo commits`  for commit notifications from a default branch.
 - `/github subscribe org/repo commits:*`  for commit notifications across all the branches.
 - `/github subscribe org/repo commits:myBranch`  for commit notifications from a specific branch.
-- `/github subscribe org/repo commits:users/*`  for commit notifications from a pattern of branches.
+- `/github subscribe org/repo commits:users/*`  for commit notifications from a pattern of users.
```
jsoref, line 141: I don't think this change is correct. I think it's just a silly example where branches are named `users/someone/something`

jsoref (changes requested):

README.md
O5ten: I see! I'll close this comment up then. :)

PR #1953, opened 2024-09-12
Updated docs for workflows filter

USSliberty:
Was not very clear that other parameters than `event` can be list too. From what I tried also `name` can be a list, not sure if there are others that can be a list too.


geramirez:
The workflow feature may no behave correctly with a list so we'll need to fix the issue before updating the docs

PR #2012, opened 2025-02-25
remove open-source claims from README

jwhitaker-gridcog:
I know there's a notice in the penultimate paragraph just after the licence, however there's also this claim in the first block. This doesn't look to have been true for 6 years now?


geramirez (approved):
thank you for the contribution

Issue #955, opened 2019-10-10
[Feature Request] Subscribe only to content that is relevant to me


gabegrand:
I've been using the GitHub Slackbot for about a month now, and I've grown frustrated with the large number of notifications about content that is not relevant to me. Even for repos with small numbers of contributors (< 10), the majority of notifications I receive are not useful to me.

**I'd like to be able to subscribe only to PRs / issues / commits / etc. that were created by me, or that I'm tagged in.** The existing filters by activity category are simply not useful for this purpose, since I care about all activity types that I'm involved in.

I believe this is a core drawback of the current Slack integration that is affecting virtually all users. There are a number of existing issues that are related:

https://github.com/integrations/slack/issues/390
https://github.com/integrations/slack/issues/878
https://github.com/integrations/slack/issues/943
https://github.com/integrations/slack/issues/773
https://github.com/integrations/slack/issues/944

In the meantime, I've turned off all notifications from the GitHub Slackbot, which is sad but necessary to keep me sane.

huehnerlady:
That is a very useful feature, I would vote for that one :)

tudoramariei:
Actually you can AND cannot do this at the moment.
You can do this...
as you can see on the page here https://github.com/integrations/slack#configuration
the extension allows you to subscribe to only some certain features (e.g. `/github subscribe owner/repository reviews,comments`)
AND
You can not use it at the moment...
it seems there's a bug and each time I'm subscribing to a repo with something like `pulls`, the extension tells me:
```
Subscribed to the following repositories
owner/repo
issues, pulls, deployments, statuses, public, commits, releases
```
the good thing is you can `/github unsubscribe owner/repo unwanted,features,list` and you're OK, you only get the notifications you need

If some more people could confirm this bug, I'll be opening it, but I think this issue can be closed

kotojo:
This would be very useful. My team has a monorepo with over 200 devs actively working out of it. Just tried to enable the integration in a private channel for myself and ended up with over 20 messages in 1 minute and had to disable it 😢

theavish:
i was just trying to figure out how to do this myself, glad to see i'm not alone

ckknight:
Since Pull Panda is no longer accepting new users, as GitHub bought them, new members of our team is missing the core functionality of Pull Reminders, which is to direct-message the developers who were assigned to a Pull Request.

Subscribing to everything is far too noisy.

G-Bro:
Commands like the below would be great :grin: 
[...]

PR #2022, opened 2025-03-28
Update README.md

leachj:
Update README to remove reference to using issues on this repository. All issues should now be raised through GitHub support

### DIFF
--- a/README.md
+++ b/README.md
@@ -746,7 +746,7 @@ If you need support or help please fill out GitHub's [Support form](https://supp
 
 | :warning: IMPORTANT          |
 |:---------------------------|
-| This repository is not accepting any code contributions. The current code which is running for GitHub and integration for Slack has significantly diverged from the code present in this repository as it contains specific code which is required to run service in GitHub infrastructure and which can not be open sourced at this point of time. **We will continue to use issues in this repository to get feedback from customers**.|
+| This repository is not accepting any code contributions. The current code which is running for GitHub and integration for Slack has significantly diverged from the code present in this repository as it contains specific code which is required to run service in GitHub infrastructure and which can not be open sourced at this point of time.|
 
 
 ## License


adaley-openai:
Hi @leachj - is there a way to access archived issues here? There is a detail in https://github.com/integrations/slack/issues/955 that I am looking for that I can no longer access. Thanks.

rmacklin:
I just left a [comment on the most recent commit](https://github.com/integrations/slack/commit/9b795556965534406f8ba299266c1e0700df02e6#commitcomment-158929782) before seeing @adaley-openai's comment here. Going to repost here, since this might be more visible than commit comments.

> I don't have any objections to the maintainers not wanting to accept issues in this repository anymore, but it's harmful to completely block access to the entire history of issues that had been discussed. All references pointing to those discussions are now broken (it even affects Google results for issues that were popular enough to get indexed).
> 
> Would you please consider re-enabling the Issues tab in this repository? You could then archive the repository to make it read-only, since you're not accepting code contributions nor issues.
> 
> Alternatively, if you don't want the repository to be archived for some reason, you could re-enable the Issues tab and set it up similar to
> https://github.com/ghostty-org/ghostty/blob/aa6c349545c1674f4fb0f0fad1c636bb034a006d/.github/ISSUE_TEMPLATE/config.yml & https://github.com/ghostty-org/ghostty/blob/aa6c349545c1674f4fb0f0fad1c636bb034a006d/.github/ISSUE_TEMPLATE/preapproved.md where the "New issue" button doesn't allow people to file blank issues and the issue template tells them not to open issues. Could even add a GitHub Actions workflow to auto-close all new issues to brute-force it.
> 
> Alternatively, would you consider exporting the contents of the issue history and making them available somewhere?
> 
> If one of these was done, the existing history of all past issue discussions & resolutions would not be hidden.
> 
> Thanks for your consideration.

@leachj @geramirez I would really appreciate your consideration. IMO, the first option seems like the best way to go - if the goal was to stop people from posting issues, and the repo had already not been accepting code contributions for years (having "significantly diverged" from the deployed code), an archived (read-only) repo with the Issues tab enabled seems to make the most sense. Is there a reason that wouldn't work?

